PR #60, opened 2014-03-17
[wip] Autocomplete for variable scope only.

rstruber:
This seems to work, but comments are certainly welcome. I am trying to write some tests, but given the nature of the feature (queuing off [TAB] and whatever is on the current buffer) it's proving difficult. Do you have suggestions?

This also seems like it could be extended a bit to start looking at class constants, application defined functions and so on. But I couldn't, just yet, figure out just how to invoke those methods and get back arrays rather than pretty print.

This also has an annoying side effect of appending a space at the end of the autocomplete.

### DIFF
--- a/src/Psy/Shell.php
+++ b/src/Psy/Shell.php
@@ -234,9 +234,9 @@ class Shell extends Application
 
         $this->readline->readHistory();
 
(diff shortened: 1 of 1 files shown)


bobthecow:
Closing in favor of #141

PR #136, opened 2015-01-07
Fixed whereami command.

hason:


src/Psy/Command/WhereamiCommand.php
```diff
@@ -100,12 +100,12 @@ protected function fileInfo()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        extract($this->fileInfo());
+        $info = $this->fileInfo();
```
hason: Yes, because ``$file`` and ``$line`` variables are not declared for IDE (magic).

bobthecow: list doesn't work unless fileInfo returns `[$file, $line]` instead of `compact('file', 'line')`

hason: @bobthecow I don't see the function ``list``.

bobthecow: http://php.net/list

blainesch:
What's the use case where this doesn't work? It appears to just be using different syntax preferences..

bobthecow:
@blainesch You mean where list doesn't work?

blainesch:
@bobthecow extract vs list. I don't see one not working and the other working... it's just a different approach?

bobthecow:
The return value of `$this->fileInfo()` is `['file' => ..., 'line' => ...]`. You can't use `list()` to get variables out of that, because it has non-numeric indices. So for `list()` to work, you'd need to change `fileInfo`'s return. That's all I was saying :)

```
Psy Shell v0.3.2 (PHP 5.6.4 — cli) by Justin Hileman
>>> $foo = ['file'=>1,'line'=>2]
=> [
       "file" => 1,
       "line" => 2
   ]
>>> list($file, $line) = $foo;
PHP error:  Undefined offset: 1 on line 1
>>> $file
PHP error:  Undefined variable: file on line 1
>>> $line
PHP error:  Undefined variable: line on line 1
>>> $bar = [1,2]
=> [
       1,
       2
   ]
>>> list($file, $line) = $bar;
=> [
       1,
       2
   ]
>>> $file
=> 1
>>> $line
=> 2
>>>
```

blainesch:
Because array's guarantee order:
`list($a, $b) = array_values(array('file' => 'foo', 'line' => 10));`

bobthecow:
Yeah, but that's so much uglier. We're the one who defines `fileInfo`, and this is the only place it's used, so why wouldn't we just change it to make it work with `list` if that's the intent?

blainesch:
My preference, unless it's actually broken, is to leave it alone.

bobthecow:
@blainesch Right now it uses `extract()`, which does need to change. This PR is a valid solution, but using `list()` would be as well. I don't have a strong preference either way, so I'm going to merge this version :)

blainesch:
@bobthecow I guess I just don't see the problem with extract?

hason:
The purpose of this PR was to solve:

1. an error message in PHP: ``PHP Notice:  Undefined offset: 3 in /psysh/src/Psy/Command/WhereamiCommand.php on line 75``
2. error messages in PHPStorm: ``Undefined variable 'file'`` and ``Undefined variable 'line'``

bobthecow:
@blainesch Like `eval`, sometimes people use `extract` for stupid things, so it bothers people. If there's a non-extract way of doing the exact same thing, which doesn't really make it harder to understand, I'm not opposed to doing it that way instead. Also, IDEs and Xdebug don't play nice with `extract` (see @hason's point 2 above).

Issue #106, opened 2014-10-24
closure problem


dlinnemeyer:
Trying some pretty strange code playing with closures and ran into this error:

```
PHP warning:  Cannot bind an instance to a static closure on /data/home/dlinnemeyer/nematode/teaching-meetings/overloading/overload.phpon line 12
```

I'm including the following file in psysh, which triggers the error:

```
<?php

$callable = function(){ echo $this->foo . "/n"; };

class Test {
    public $foo;
}

$test = new Test();
$test->foo = "bar";

$bound_callable = $callable->bindTo($test);

$bound_callable();
```

Also, I get these error messages on exit from psysh:

```
PHP Warning:  unserialize(): Function spl_autoload_call() hasn't defined the class it was called for in /data/home/dlinnemeyer/psysh/src/Psy/ExecutionLoop/ForkingLoop.php on line 64
PHP Stack trace:
PHP   1. {main}() /data/home/dlinnemeyer/psysh/bin/psysh:0
PHP   2. call_user_func() /data/home/dlinnemeyer/psysh/bin/psysh:100
PHP   3. {closure:/data/home/dlinnemeyer/psysh/bin/psysh:40-100}() /data/home/dlinnemeyer/psysh/bin/psysh:100
PHP   4. Psy\Shell->run() /data/home/dlinnemeyer/psysh/bin/psysh:99
PHP   5. Symfony\Component\Console\Application->run() /data/home/dlinnemeyer/psysh/src/Psy/Shell.php:215
PHP   6. Psy\Shell->doRun() /data/home/dlinnemeyer/psysh/vendor/symfony/console/Symfony/Component/Console/Application.php:124
PHP   7. Psy\ExecutionLoop\ForkingLoop->run() /data/home/dlinnemeyer/psysh/src/Psy/Shell.php:243
PHP   8. unserialize() /data/home/dlinnemeyer/psysh/src/Psy/ExecutionLoop/ForkingLoop.php:64
```

The example works fine when run through php itself.

bobthecow:
I think I figured out what's happening. I'm just not sure how to work around it :)

Every bit of code you run in PsySH is executed inside a closure. This includes any files you include or require. See, for example, [this line](https://github.com/bobthecow/psysh/blob/master/src/Psy/ExecutionLoop/Loop.php#L45), where the files passed in from the command line are included (e.g. `psysh test.php`).

[...]

PR #138, opened 2015-01-16
Timer command for function time profiling

Markcial:
A timer command for function time profiling.

Usage : timeit "closure or code to execute"

screenshot from 2015-01-16 16 59 052

### DIFF
--- a/src/Psy/Command/TimeitCommand.php
+++ b/src/Psy/Command/TimeitCommand.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Psy\Command;
+
+use Psy\Configuration;
+use Psy\Shell;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * Class TimeitCommand
+ * @package Psy\Command
+ */
+class TimeitCommand extends Command
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('timeit')
+            ->setDefinition(array(
+                new InputArgument('target', InputArgument::REQUIRED, 'A target object or primitive to profile.', null),
+            ))
+            ->setDescription('Profiles with a timer.')
+            ->setHelp(
+                <<<HELP
+Time profiling for functions and commands.
+
+e.g.
+<return>>>> timeit \$closure()</return>
+HELP
+            );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $target = $input->getArgument('target');
+        $start = microtime(true);
+
+        /** @var Shell $shell */
+        $shell = $this->getApplication();
+        $sh = new Shell(new Configuration());
+        $sh->setOutput($output);
+        $sh->setScopeVariables($shell->getScopeVariables());
+        $sh->execute($target);
+
+        $end = microtime(true);
+
+        $output->writeln(sprintf('<info>Command took %.6f seconds to complete.</info>', $end-$start));
+    }
+}
--- a/src/Psy/Command/TimeitCommand.php
+++ b/src/Psy/Command/TimeitCommand.php
@@ -6,7 +6,6 @@ use Psy\Configuration;
 use Psy\Shell;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Input\InputArgument;
 
 /**
  * Class TimeitCommand
@@ -21,10 +20,8 @@ class TimeitCommand extends Command
     {
         $this
             ->setName('timeit')
-            ->setDefinition(array(
-                new InputArgument('target', InputArgument::REQUIRED, 'A target object or primitive to profile.', null),
-            ))
             ->setDescription('Profiles with a timer.')
+            ->setUsesWholeStringInput(true)
             ->setHelp(
                 <<<HELP
 Time profiling for functions and commands.
@@ -40,16 +37,16 @@ HELP
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $target = $input->getArgument('target');
-        $start = microtime(true);
+        $target = $input->getLine();
 
         /** @var Shell $shell */
         $shell = $this->getApplication();
         $sh = new Shell(new Configuration());
         $sh->setOutput($output);
         $sh->setScopeVariables($shell->getScopeVariables());
-        $sh->execute($target);
 
+        $start = microtime(true);
+        $sh->execute($target);
         $end = microtime(true);
 
         $output->writeln(sprintf('<info>Command took %.6f seconds to complete.</info>', $end-$start));
--- a/src/Psy/ExecutionLoop/Loop.php
+++ b/src/Psy/ExecutionLoop/Loop.php
@@ -38,7 +38,7 @@ class Loop
      */
     protected function getExecutionClosure()
     {
-        return function (Shell &$__psysh__, $code = null) {
+        return function (Shell &$__psysh__) {
 
             try {
                 extract($__psysh__->getScopeVariables());
@@ -47,7 +47,7 @@ class Loop
                 ob_start();
 
                 set_error_handler(array($__psysh__, 'handleError'));
-                $_ = eval(!is_null($code) ? $code : $__psysh__->flushCode());
+                $_ = eval($__psysh__->flushCode());
                 restore_error_handler();
 
                 $__psysh_out__ = ob_get_contents();
@@ -58,11 +58,8 @@ class Loop
 
                 // a bit of housekeeping
                 unset($__psysh_out__);
-                unset($code);
 
                 $__psysh__->setScopeVariables(get_defined_vars());
-            } catch (BreakException $_e) {
-                throw $_e;
             } catch (\Exception $_e) {
                 restore_error_handler();
                 if (ob_get_level() > 0) {
@@ -117,8 +114,9 @@ class Loop
         if (self::bindLoop()) {
             $executionClosure = $this->setClosureShellScope($shell, $executionClosure);
         }
+        $shell->addCode($code);
 
-        return $executionClosure($shell, $code);
+        return $executionClosure($shell);
     }
 
     /**
@@ -165,7 +163,7 @@ class Loop
             return $loop->bindTo($that, get_class($that));
         }
 
-        return $loop->bindTo(null);
+        return $loop->bindTo(null, null);
     }
 
     /**
--- a/src/Psy/ExecutionLoop/Loop.php
+++ b/src/Psy/ExecutionLoop/Loop.php
@@ -20,6 +20,9 @@ use Psy\Exception\BreakException;
  */
 class Loop
 {
+    /** @var \Psy\CodeCleaner  */
+    protected $cleaner;
+
     /**
      * The non-forking loop doesn't have much use for Configuration :)
      *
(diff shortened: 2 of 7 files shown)


src/Psy/Command/TimerCommand.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    protected function runTarget($target)
+    {
+        /** @var Shell $shell */
+        $shell = $this->getApplication();
+        $sh = new \Psy\Shell(new Configuration());
+        $sh->setScopeVariables($shell->getScopeVariables());
+        $sh->addCode($target);
+        $shell->writeReturnValue(eval($sh->flushCode()));
```
bobthecow: This should probably be run in `$sh`, not `eval()` here, right?

bobthecow:
Nice!

src/Psy/Command/TimerCommand.php
```diff
@@ -0,0 +1,60 @@
...
+use Psy\Shell;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * Class TimerCommand
+ * @package Psy\Command
+ */
+class TimerCommand extends Command
```
bobthecow: Class name and command should match.

src/Psy/Command/TimerCommand.php
Markcial: Do you mind giving me a hint about the proper command for running the code inside the shell? I tried some things and i ended doing the eval directly. I will review the code again to see what i can find, but if you know the exact command would be nice. Thanks

bobthecow:
There's currently no way to eval arbitrary code from outside a shell instance, but it shouldn't be hard to add:

```diff
diff --git a/src/Psy/ExecutionLoop/Loop.php b/src/Psy/ExecutionLoop/Loop.php
index bf281cd..8220105 100644
--- a/src/Psy/ExecutionLoop/Loop.php
+++ b/src/Psy/ExecutionLoop/Loop.php
@@ -135,4 +135,51 @@ class Loop
     {
         // no-op
     }
+
+    public function evaluate(Shell $shell, $code)
+    {
+        $closure = function($__psysh__, $__psysh_code__) {
+            extract($__psysh__->getScopeVariables());
+
+            try {
+                // evaluate the supplied code
+                ob_start();
+
+                set_error_handler(array($__psysh__, 'handleError'));
+                $_ = eval($__psysh_code__);
+                restore_error_handler();
+
+                $__psysh_out__ = ob_get_contents();
+                ob_end_clean();
+
+                $__psysh__->writeStdout($__psysh_out__);
+                return $_;
+            } catch (\Exception $_e) {
+                restore_error_handler();
+                if (ob_get_level() > 0) {
+                    ob_end_clean();
+                }
+                $__psysh__->writeException($_e);
+            }
+        };
+
+        // bind the closure to $this from the shell scope variables...
+        // skip binding on HHVM, see https://github.com/facebook/hhvm/issues/1203
+        if (version_compare(PHP_VERSION, '5.4', '>=') && !defined('HHVM_VERSION')) {
+            $that = null;
+            try {
+                $that = $shell->getScopeVariable('this');
+            } catch (\InvalidArgumentException $e) {
+                // well, it was worth a shot
+            }
+
+            if (is_object($that)) {
+                $loop = $loop->bindTo($that, get_class($that));
+            } else {
+                $loop = $loop->bindTo(null);
+            }
+        }
+
+        return $closure($shell, $code);
+    }
 }
diff --git a/src/Psy/Shell.php b/src/Psy/Shell.php
index e3171de..92fe0c2 100644
--- a/src/Psy/Shell.php
+++ b/src/Psy/Shell.php
@@ -244,6 +244,14 @@ class Shell extends Application
         $this->loop->run($this);
     }
 
+    public function evaluate($code)
+    {
+        $code = $this->cleaner->clean(array($code));
+
+        // Wheee!
+        return $this->loop->evaluate($this, $code);
+    }
+
     /**
      * Read user input.
      *
```

Using that — or something like that, but actually tested :) — you can call `$shell->evaluate($code)` and it'll return the result.

Markcial:
Yeah i thought about adding a method for the loop class, but i thought it would be a little bit overkill. but now that i know that you agree about this I will do the evaluate function on the loop class and use it inside the command. Just give me a couple of days because the testing and stuff. Thanks

bobthecow:
There have been several times in the past that I've almost implemented this for other commands, so it's probably not overkill.

Markcial:
Another way to solve this could be making a public api to the beforeLoop and the afterLoop, this way the commands will have access to add their own routines, for example adding a tick with the time, or maybe benchmarking memory usage, etc

bobthecow:
You mean rather than starting a sub-shell? The downside to that is that it would modify _this_ shell's state. The last error, the return value, the current variable state, etc would all be changed when you were just "benchmarking" something… it might make sense to have explicit side effects possible, but it would be weird for it to extend to things like the return value and last error, IMO.

Markcial:
You are completely right, you know the psysh codebase better than me and i was not thinking on the downsides. Netx week i will push the experiment with the subshell and the evaluate method. I will try to add some testing too for that evaluate method.

src/Psy/Command/TimeitCommand.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $target = $input->getArgument('target');
+        $start = microtime(true);
+
+        /** @var Shell $shell */
+        $shell = $this->getApplication();
+        $sh = new Shell(new Configuration());
+        $sh->setOutput($output);
+        $sh->setScopeVariables($shell->getScopeVariables());
+        $sh->execute($target);
```
bobthecow: let's move the `$start` to right before running `$sh->execute()` so that we have as little overhead as possible here.

src/Psy/Command/TimeitCommand.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('timeit')
+            ->setDefinition(array(
+                new InputArgument('target', InputArgument::REQUIRED, 'A target object or primitive to profile.', null),
```
bobthecow: I've been thinking about this a bit, and the more I consider it the more I feel that we need a different input string parsing type.

Right now, all command input is parsed like command line args. We would really be better served by a hybrid between command line args and "just give me the string they typed". Here, for example, if you're trying to time `foo($bar, 'baz')` you'll run into problems, because InputArgument will just grab part of that as `target` :-/

src/Psy/ExecutionLoop/Loop.php
```diff
@@ -27,17 +30,55 @@ class Loop
...
+            }
+        };
+    }
+
+    /**
+     * @return callable
+     */
+    protected function getLoopClosure(\Closure $executionClosure)
+    {
+        return function (Shell &$__psysh__) use ($executionClosure) {
```
bobthecow: This will put `$executionClosure` in the local scope while running the REPL. I've been very careful to _only_ expose `$__psysh__` so that the amount of unexpected PsySH stuff "leaking" into the REPL is very limited. And then I added a CodeCleaner pass to make it hard for you to even touch that one variable :P

Is there a way to do this without exposing another variable name?

src/Psy/ExecutionLoop/Loop.php
```diff
@@ -50,67 +91,81 @@ public function run(Shell $shell)
...
+            $that = $shell->getScopeVariable('this');
+        } catch (\InvalidArgumentException $e) {
+            // well, it was worth a shot
+        }
+        if (is_object($that)) {
+            return $loop->bindTo($that, get_class($that));
         }
 
-        $loop($shell);
+        return $loop->bindTo(null);
```
bobthecow: This is a regression, it reverts the fix for #106

src/Psy/ExecutionLoop/Loop.php
```diff
@@ -27,17 +30,55 @@ class Loop
...
-     * Run the execution loop.
-     *
-     * @param Shell $shell
+     * @return callable
      */
-    public function run(Shell $shell)
+    protected function getExecutionClosure()
     {
-        $loop = function ($__psysh__) {
+        return function (Shell &$__psysh__, $code = null) {
```
bobthecow: Again, we're leaking a variable into local scope. This one's even more problematic, because it's *way* more likely that someone would have a variable called $code, which would be extracted on line 44 below and executed :-/

It's really looking like having a second code path just for `execute()` would be far cleaner.

src/Psy/ExecutionLoop/Loop.php
```diff
@@ -27,17 +30,55 @@ class Loop
...
+
+                $__psysh__->writeStdout($__psysh_out__);
+                $__psysh__->writeReturnValue($_);
+
+                // a bit of housekeeping
+                unset($__psysh_out__);
+                unset($code);
+
+                $__psysh__->setScopeVariables(get_defined_vars());
+            } catch (BreakException $_e) {
```
bobthecow: For example, handling BreakException only makes sense if you could be executing commands, not if you're evaluating code.

src/Psy/Command/TimeitCommand.php
Markcial: I will need to add a additional class, for this, but it is possible

src/Psy/ExecutionLoop/Loop.php
Markcial: the executionClosure scope is not visible, because is the closure that has the context of the call, the issue related to the code variable is in fact leaked an could drive to errors, but this one has no visibility, i might better came along with a better solution for the code variable anyways, i am trying get a working solution for that.

src/Psy/ExecutionLoop/Loop.php
Markcial: do we have a test for that, or can you explain me how to add it to the tests? thanks

src/Psy/ExecutionLoop/Loop.php
bobthecow: Right, I didn't notice that there was another level of closure when I made that comment :)

src/Psy/ExecutionLoop/Loop.php
bobthecow: There is not a test for that. I'll take a look.

src/Psy/Input/WholeStringInput.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Psy\Input;
+
+use Symfony\Component\Console\Input\InputDefinition;
+use Symfony\Component\Console\Input\StringInput;
+
+/**
+ * Class WholeStringInput
+ * @package Psy\Input
+ */
+class WholeStringInput extends StringInput
```
bobthecow: :+1: This seems like a good approach to the issue.

It could also be extended in the future if we have a whole string input command that uses flags, so we could do something like `timeit --verbose foo()` and be able to process options as well. But not yet :)

src/Psy/ExecutionLoop/Loop.php
bobthecow: Since there is now an execution closure, does the loop closure still need to be a closure? Could we move all of this mess into the run method itself, and save ourselves from the weird variables and convoluted interaction here?

If we do, we would probably have to do something about the user-defined includes, because they'd still want to be executed in a closure…

This isn't a request to change anything yet, I'm not even sure if it's a good idea. I'm just asking what you think :)

Markcial: I suppose it doesn't make a big difference, i did the closure in a separated function for a cleaner method calls, but being the main loop closure it doesn't matter if it is created as a lambda or received as a parameter from a function, so my vote is to have it in a sepparated function for a cleaner structure, but it is not relevant so if you want it in the loop call it wouldnt be a problem at all and it wouldnt take more than a minute to commit. TL;DR as you decide.

Markcial:
Ping! do we have any news about this ticket? i know that the loop has a heavy rewrite that needs good review, but the new loop maybe could help me with this ticket => https://github.com/bobthecow/psysh/issues/154 i have a functional experiment, but i think that the current quirky code would be a lot more cleaner with the splitted functions made in this loop.

bobthecow:
Sorry, I haven't had time to do a more in-depth review yet, and you're absolutely right that it'll take a bit more. To be honest, I've got a bit of a crazy week or two ahead of me and don't see getting this shipped before then.

Right now I'm thinking that I'd like to get `v0.4.0` out the door (asap) with autocomplete and everything currently in the develop branch, and get the loop refactor, etc into `v0.5.0`.

Markcial:
I supposed so, then if you don't mind i will do the quirky PR with the `CancellableForkingLoop` as a different class so anyone can test it without loosing the normal `ForkingLoop` linking here with a note explaining there is a candidate for the shell loop refactor, so anyone whom might be interested can review and have a peek to this slightly hidden piece of code in the non-related issue name.

GrahamCampbell:
Any news on this?

Markcial:
I am waiting for a code review. Sorry

damiankloip:
@bobthecow what do we need to do to get this done? This is a very handy feature!

bobthecow:
Oh wow. Sorry it took so long to get into a release.

This has been merged into the `feature/execute` branch, which will (barring any issues) be merged into `develop` shortly and released with v0.9.x.

PR #141, opened 2015-01-21
Tab completion support added

Markcial:
With this PR the config now has a parameter tabCompletion that enables tab completion support for php variables, functions and class constructors.


src/Psy/Shell.php
```diff
@@ -71,6 +73,11 @@ public function __construct(Configuration $config = null)
         $this->context  = new Context();
         $this->includes = array();
         $this->readline = $this->config->getReadline();
+        if ($this->config->getTabCompletion()) {
+            $this->completion = new TabCompletion();
+            $this->completion->setContext($this->context);
```
bobthecow: I was going to comment that I'd prefer injecting dependencies via the constructor, but it looks like you've already written it that way :) Let's use that here?

bobthecow:
:+1:!

src/Psy/Shell.php
Markcial: in just a minute

src/Psy/Readline/TabCompletion.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        readline_completion_function(array(&$this, 'callback'));
+    }
+
+    /**
+     * @param $input
+     * @param $index
+     * @return array
+     */
+    public function callback($input, $index)
```
bobthecow: I'd love to be able to test this... Can we extract the guts of `callback()` into a function that takes `$input, $index, readline_info())` so that we can pass in fake data?

bobthecow:
Thanks for taking this on :)

src/Psy/Readline/TabCompletion.php
Markcial: ah ok, it wont be much trouble, just give me a moment

src/Psy/Readline/TabCompletion.php
```diff
@@ -0,0 +1,65 @@
...
+        $info = readline_info();
+
+        // the char just before the cursor is a dollar? send a variable context
+        $charAt = substr($info['line_buffer'], $index-1, 1);
+        if ($charAt == '$') {
+            return array_keys($this->context->getAll());
+        }
+
+        $parenthesize = function ($name) {
+            return sprintf('%s()', $name);
```
bobthecow: Would it make sense to check the arity of the function being completed and only add the closing paren if it's a function that doesn't accept args?

bobthecow: (possibly caching that info on this object to keep from looking it up all the time)

Markcial: Yes, It could be done with reflection, but as you said, the objects might need to be cached because the reflected entities would be spending huge amounts of memory if they are not stored in a certain reference. I will dig on that anyways, but it will take me a couple of days more maybe

src/Psy/Readline/TabCompletion.php
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+
+        // for everything else send the functions names
+        $functions = get_defined_functions();
+        return array_map(
+            $parenthesize,
+            array_merge(
+                $functions['internal'],
+                $functions['user']
```
bobthecow: We should add fake functions (control structures or keywords that masquerade as functions) here as well. Prolly store them in a static var on this class?

```
['declare', 'print', 'require', 'require_once', 'import', 'import_once', 'echo', 'empty']
```

… I'm sure I'm missing a few, but that's a start :)

I don't think we need to go as far as including `switch`, `if`/`else`, the various sorts of loops, etc, because those have a bunch of curly brace baggage. But we should include the ones that look and act like functions.

src/Psy/Readline/TabCompletion.php
bobthecow: It doesn't have to be in v1 of the feature.

src/Psy/Readline/TabCompletion.php
Markcial: i can do a copypasta from [this comment](http://php.net/manual/en/reserved.keywords.php#110244) on the php site and then send to the completion callback if the input matches

src/Psy/Readline/TabCompletion.php
Markcial: I supposed, i was only preparing the files for the tests and checking the errors from the travis bot, i will resubmit code in a moment

src/Psy/Readline/TabCompletion.php
bobthecow: Nice! Of those, it seems like these would make the most sense:

```
'array'
'clone'
'declare'
'die'
'echo'
'empty'
'eval'
'exit'
'include'
'include_once'
'isset'
'list'
'print'
'require'
'require_once'
'unset'
```

Those are the most function-like of them, and the ones that I would be most likely to expect to have completed. There are a few more that are function-like but don't make much sense in a REPL context, and the rest of them are either complex control structures, or things that would need to be completed without parens, so let's hold off on them.

Markcial:
I added a phpunit test case with a simple test, so you have a quickstart for testing the tabcompletion class, i have runned out of time for today, but maybe for next week i will add some of the ideas you submitted in the thread

src/Psy/Readline/TabCompletion.php
```diff
@@ -0,0 +1,65 @@
...
+        $charAt = substr($info['line_buffer'], $index-1, 1);
+        if ($charAt == '$') {
+            return array_keys($this->context->getAll());
+        }
+
+        $parenthesize = function ($name) {
+            return sprintf('%s()', $name);
+        };
+
+        if (strlen($info['line_buffer']) > 4) {
```
bobthecow: potential for bugs here: `$info['line_buffer']` seems to be quite literally the buffer they use for input, and it often contains things that aren't part of the current line (e.g. left over from the last line). We should do `$line = substr($info['line_buffer'], 0, $info['end'])` somewhere early on in this function and always refer to that instead.

Markcial: ok, and it can be tested too then

Markcial:
I am sorry for the failed tests, but i don't know what does the PHPCS means with that error code, i will dig a little bit more, but if you have any input about this ping me here so i can solve the failed tests, thanks

bobthecow:
Yeah, I'm not sure. It's not your file, though, so it's not your problem :)

Markcial:
Hi Justin, seems to be that there is an open discussion about the conditional that breaks the PSR-2 specs and makes the CodeSniffer warn about an error https://github.com/squizlabs/PHP_CodeSniffer/pull/188 the issue seems to be closed and merged, so maybe an update of the library will fix this.

Markcial:
Unfortunatelly i tried the latest version of CodeSniffer (the one you were already using) and this patch is too recent to be applied, it might be avaliable for the next release. I found a workaround for that and is assigning a boolean for the multiline condition and then put that boolean in the enclosed condition, if you wan't the tests until CS is updated i can rewrite the condition on that file so the TravisBot can finally mark the tests as succesful

bobthecow:
I like the autocomplete matchers approach. This could make it possible for things like ezsh (ezComponents shell, based on PsySH) to add their own matchers for their own crazy model stuff :)

Markcial:
Glad you liked it, i will work a little bit more on this throught the weekend, i know that currently all the code is a little bit messy and hacky, just fiddling. I expect to have cleaned the mess and write tests too for the next weekend.

Markcial:
I am still working on it, i found that the readline has some quirks with the namespaces, if the user inputs a namespaced class like `Foo\Bar\Spam:`: the input only receives the non namespaced part, `Spam::` it adds another bit of complexity but i will deal with it. There is another thing that will be a nice to have, reflect the classes or objects to detect if they accept params or not, in case they accept params autocomplete with open parenthesis and add the coma after the param. So it will take a little bit of time too. I will keep yo up to date when i had completed some more code and tests.

bobthecow:
Cool. Thanks!

src/Psy/Configuration.php
```diff
@@ -772,6 +774,26 @@ public function getLoop()
     }
 
     /**
+     * Enables tab completion mode
+     *
+     * @param $tabCompletion
+     */
+    public function setTabCompletion($tabCompletion)
+    {
+        $this->tabCompletion = (bool) $tabCompletion;
```
bobthecow: We should check whether the Readline implementation supports tab completion rather than setting one up and having it just sit there :)

bobthecow: (sorry, meant to leave that comment on `getTabCompletion`, not `setTabCompletion`)

src/Psy/Shell.php
```diff
@@ -74,6 +76,11 @@ public function __construct(Configuration $config = null)
 
         parent::__construct('Psy Shell', self::VERSION);
 
+        if ($this->config->getTabCompletion()) {
+            $this->completion = new AutoCompleter($this->context, $this->all());
+            $this->completion->activate();
+        }
```
bobthecow: I think I'd prefer to move instantiation of this service to Configuration, where other services are instantiated. That way users can interact with it in the psysh config file, much the same way they can interact with commands, presenters and other extension points.

Then I could return this from my config:

```php
return [
  'tabCompletionMatchers' => [
    new EloquentModelMatcher(),
    new FancyPantsMatcher()
  ]
]
```

… and get some custom matchers that meet my needs, to go with the default set.

Along with that we would probably have to change to a ContextAwareInterface style dependency injection for the matchers that depend on context, shell, etc, so that users wouldn't have to have access to those things when setting up the initial configuration. Commands already do this (see ReflectingCommand, WtfCommand).

Maybe we use the config as an autocompleter factory, and just activate it here?

```php
if ($this->config->getTabCompletion()) {
  $this->completion = $this->config->getAutoCompleter();
  $this->completion->activate();
}
```

I haven't thought through this completely, but it seems like a direction we should head :)

src/Psy/TabCompletion/AutoCompleter.php
```diff
@@ -0,0 +1,114 @@
...
+        $commandsMatcher->setCommands(
+            array_map(function (Command $command) {
+                return $command->getName();
+            }, $commands)
+        );
+    }
+
+    public function activate()
+    {
+        readline_completion_function(array(&$this, 'callback'));
```
bobthecow: do we need a deactivate? if we exit a psysh debugging session and end up back in a regular `php -a` or something, will this screw everything up?

test/Psy/Test/TabCompletion/AutoCompleterTest.php
```diff
@@ -0,0 +1,66 @@
...
+            array('s', 'stdClass'),
+            array('s', 'stdClass'),
+            array('array_', 'array_search'),
+            array('$bar->', 'load'),
+            array('$b', 'bar' ),
+            array('6 + $b', 'bar' ),
+            array('$f', 'foo' ),
+            array('l', 'ls'),
+            array('sho', 'show'),
+            array('12 + clone ', 'foo' ),
```
bobthecow: should this be completing with a `$` since there isn't one in the input yet?

bobthecow:
I can't wait to have this. What's your take on the current status?

Markcial:
What do you mean with "what's your take on the current status". I am a non native english speaker sorry. Do you want my opinion on your thoughts or my status on the code? If you meant my opinion, i think that you are completelly right, the autocompleter has no means if you cannot add your own completers because of personallized code like symfony, code igniter, etcetera. But if the autocompleter needs to be done this way i need to spend a little bit more time, because we will need some interfaces for contractual reasons when the end users write their own matchers, and maybe we need to allow them to add the rulers for better matching capabilities. About the context, you are right too, my bad, i must had used the contextaware pattern, i will rewrite the matchers for that and use a factory, so the autocompleter can get their dependencies.  BTW i found a weird behaviour in readline with the backslash, i need to dig a bit about this. With all that pending stuff you have commented here, i might be done in a week or two, depending on my spare time.

test/Psy/Test/TabCompletion/AutoCompleterTest.php
Markcial: the readline ignores inputs like dollar or backslash, is a weird behaviour that i am currently digging about.

src/Psy/TabCompletion/AutoCompleter.php
Markcial: yes, you are right, i will add the deactivate on __destruct

src/Psy/Configuration.php
Markcial: is ok for me, with module_loaded('readline') is enough?

src/Psy/TabCompletion/AutoCompleter.php
Markcial: after a bit of testing seems to be that is not necessary to uninstall the completion function, i will skip this step.

test/Psy/Test/TabCompletion/AutoCompleterTest.php
bobthecow: Maybe we should not complete variables until they type a `$` then, for now? So:

 * `12 + clone [tab][tab]` wouldn't include variable names
 * `12 + clone $[tab][tab]` would include variable names

Or is it weird about leaving the `$` that they already typed?

src/Psy/TabCompletion/AutoCompleter.php
bobthecow: okay.

bobthecow: oh, did you try it with pcntl (forking loop) disabled? that might make a difference, too, depending on how readline completions deal with forks.

src/Psy/Configuration.php
bobthecow: :+1:

bobthecow:
Sorry I wasn't completely clear. Yes, I was looking for a high level opinion on the state of the code and this feature.

The autocompleter doesn't need to be perfect on the first pass... The CodeCleaner went through a couple of releases before it was more than just an ugly hack. Presenters used to be hardcoded, but now are extensible. It's better to have _any_ autocompleter than to wait until we have the perfect autocompleter :)

That said, it'd be nice to have something extensible. So if it's a small amount of work, we should look into that. If it's a large amount of work, maybe we hold off. I was mostly asking about that because if we need to fundamentally rethink anything about the architecture to allow it, it'd be nice to know that beforehand, even if we don't change it now. If we might need to make a backwards compatibility break to the public api or config options, we should think that through now as well.

As far as readline weird behaviors, I hear you. It's a pretty awkward API, and it doesn't always do what it seems like it should :)

Maybe it would be best to make sure the architecture, interfaces, configuration and implementation are where we want them, even if we can't complete everything right away. For example, if namespaced things don't work, we could ship completion for variables and non-namespaced classes, functions, and constants first. Any completion is better than the current completion, so let's not focus too much on getting all of the matchers implemented. We can ship the first iteration with some matchers disabled if we need to.

Make sense?

src/Psy/Configuration.php
bobthecow: `extension_loaded('readline')` obvs ;)

src/Psy/TabCompletion/AutoCompleter.php
Markcial: I will test that in a docker VM. but i will put that in the bottom list for the TODO.

test/Psy/Test/TabCompletion/AutoCompleterTest.php
Markcial: I am trying to do as you say, i am trying by splitting the code by tokens in a lexer non strict way with token_get_all function, i came out with a solution named rulers that i am currently removing because i do not like to have that huge amount of files for simple rules. I will push a new set of files by the end of the day so you can check how it works, but mainly is like you say, is a little bit weird because the completion shows the variables without the dollars.

Markcial:
I agree, expect a PR in a couple of hours, i did some rewrite and made it configurable

src/Psy/TabCompletion/AutoCompleter.php
bobthecow: Don't worry too much about it. I'll take a look in a bit.

Markcial:
basic functionality is here, but anyways needs more testing, still need to reflect the methods looking if parameters are needed, add more complex tests, etcetera, but basic completion works, i will keep adding more functionalities and explain in a brief doc how to add your own matchers.

src/Psy/Configuration.php
```diff
@@ -772,6 +775,34 @@ public function getLoop()
...
+    {
+        return function_exists('readline_info');
+    }
+
+    /**
+     *
+     */
+    public function getAutoCompleter()
+    {
+        $completer = new AutoCompleter();
```
bobthecow: this should store the completer instance on `$this->completer` like config does with all other services.

src/Psy/Configuration.php
```diff
@@ -220,7 +223,7 @@ public function loadConfig(array $options)
             }
         }
 
-        foreach (array('commands') as $option) {
+        foreach (array('commands', 'tabCompletionMatchers') as $option) {
```
bobthecow: `addTabCompletionMatchers` hasn't been implemented yet?

src/Psy/Shell.php
```diff
@@ -203,6 +218,34 @@ protected function getDefaultCommands()
     }
 
     /**
+     * @return array
+     */
+    protected function getTabCompletionMatchers()
+    {
+        $this->addTabCompletionMatchers(
```
bobthecow: maybe add an `if (empty($this->tabCompletionMatchers))` guard here? I know it's protected, so we don't need to worry about random users calling it a second time, but that'll help guard against future changes doing something dumb and re-initializing these :)

src/Psy/TabCompletion/AutoCompleter.php
```diff
@@ -0,0 +1,68 @@
...
+        $matches = array();
+        foreach ($this->matchers as $matcher) {
+            if ($matcher->hasMatched($tokens)) {
+                $matches = array_merge($matcher->getMatches($tokens), $matches);
+            }
+        }
+
+        $matches = array_unique($matches);
+
+        return !empty($matches) ? $matches : array('');
```
bobthecow: returning an empty string here keeps it from autocompleting file names and dumb things like that, yeah?

src/Psy/TabCompletion/Matchers/ObjectAttributesMatcher.php
```diff
@@ -0,0 +1,52 @@
...
+        $input = $this->getInput($tokens);
+
+        $firstToken = array_pop($tokens);
+        if (self::tokenIs($firstToken, self::T_STRING)) {
+            // second token is the object operator
+            array_pop($tokens);
+        }
+        $objectToken = array_pop($tokens);
+        $objectName = str_replace('$', '', $objectToken[1]);
+        $object = $this->context->get($objectName);
```
bobthecow: should be using `$this->getVariable`

src/Psy/TabCompletion/Matchers/ObjectMethodsMatcher.php
```diff
@@ -0,0 +1,48 @@
...
+        $input = $this->getInput($tokens);
+
+        $firstToken = array_pop($tokens);
+        if (self::tokenIs($firstToken, self::T_STRING)) {
+            // second token is the object operator
+            array_pop($tokens);
+        }
+        $objectToken = array_pop($tokens);
+        $objectName = str_replace('$', '', $objectToken[1]);
+        $object = $this->context->get($objectName);
```
bobthecow: `$this->getVariable` again :)

src/Psy/TabCompletion/Matchers/AbstractMatcher.php
```diff
@@ -0,0 +1,166 @@
...
+namespace Psy\TabCompletion\Matchers;
+
+use Psy\Context;
+use Psy\ContextAware;
+
+/**
+ * Class AbstractMatcher
+ * @package Psy\TabCompletion\Matchers
+ */
+abstract class AbstractMatcher implements ContextAware
```
bobthecow: It looks like only three matchers actually need ContextAware (ObjectAttributes, ObjectMethods, and Variables). Let's extend this with a ContextAwareMatcher (or AbstractContextAwareMatcher, I don't care which name) for them to extend so that only those three matchers are actually ContextAware.

bobthecow:
:+1: this is looking really good!

src/Psy/TabCompletion/Matchers/AbstractMatcher.php
```diff
@@ -0,0 +1,166 @@
...
+{
+    /** Syntax types */
+    const CONSTANT_SYNTAX = '^[A-Z0-9-_]+$';
+    const VAR_SYNTAX = '^\$[a-zA-Z0-9-_]+$';
```
Taluu: How about using the [real regex](http://php.net/manual/en/language.variables.basics.php), `[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*` ?

src/Psy/TabCompletion/AutoCompleter.php
Markcial: yes sorry, i was tired of file path completions, but maybe some day with a matcher could be added too.

src/Psy/TabCompletion/Matchers/AbstractMatcher.php
Markcial: if you are referring to this note : 
> Note: For our purposes here, a letter is a-z, A-Z, and the ASCII characters from 127 through 255 (0x7f-0xff).

It means the **content** of the variable or constant, not the name, the real regex is not like this, or you will end getting constant names as weird as "T�OT�", being all this chars in this contant inside the ranges defined in the note,but definitively are not a valid constant name.

Taluu: Nope, this is a valid name...

> Variable names follow the same rules as other labels in PHP. A valid variable name starts with a letter or underscore, followed by any number of letters, numbers, or underscores. As a regular expression, it would be expressed thus: '[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*'

I'm refering to the text before the note

Taluu: and a constant is not always caps (it is a standard, yes, but still, `define('foo', 'bar'); foo;` **is** valid

Markcial: Okay, so you **really** think that chars like this are valid for a variable or constant name?
<pre>
>>> print '\x5f'
_
>>> print '\x5d'
]
>>> print '\x5c'
\
>>> print '\x5e'
^
</pre>
I can trust the docs until some point, but this is certainly confusing.

Taluu: Well, the spec also says so : https://github.com/php/php-langspec/blob/master/spec/09-lexical-structure.md#names

Markcial: I will add the regular expression because the docs, but some of the chars in this range are whitespaces or non legible chars.

src/Psy/TabCompletion/AutoCompleter.php
bobthecow: no, that's fine. i was just confirming the reason for the empty string.

i agree, we should do this. there are only a few cases where completing filenames makes sense. we can have a matcher specifically for that one of these days :)

src/Psy/TabCompletion/Matchers/AbstractMatcher.php
```diff
@@ -0,0 +1,143 @@
...
+    public function hasMatched(array $tokens)
+    {
+        return false;
+    }
+
+    /**
+     * @param  $tokens
+     * @return string
+     */
+    protected function getInput(array $tokens)
```
bobthecow: should this the static?

src/Psy/TabCompletion/Matchers/AbstractMatcher.php
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        return $var;
+    }
+
+    /**
+     * @param $tokens
+     * @return string
+     */
+    protected function getNamespaceAndClass($tokens)
```
bobthecow: should this be static as well?

src/Psy/TabCompletion/Matchers/AbstractMatcher.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        return preg_match(sprintf('#^%s#', $prefix), $word);
+    }
+
+    /**
+     * @param $token
+     * @param  string $syntax
+     * @return bool
+     */
+    public static function hasSyntax($token, $syntax = self::VAR_SYNTAX)
```
bobthecow: It looks like we're not currently using this anywhere.

bobthecow:
@Markcial This looks like it's basically good to merge. Are we waiting on anything else?

Markcial:
maybe some docs, but for eveything else i think that it has decent features for basic tab completion.

bobthecow:
OK, let's get this shipped!

I updated, fixed and added docblocks here:

https://github.com/bobthecow/psysh/compare/develop...feature/tab-completion

I also had to change config just a bit so that specifying custom matchers would actually work. In the process, I wrote two custom matchers (for MongoDB) but they're disabled by default. This way there's even an example to add to the README section about adding custom matchers :)

@Markcial I added license headers to the new files in f66804d. We don't have a formal CLA, but it'd be great to get a :+1: indicating you're okay with this :)

bobthecow:
Also, after I merge this, I need to update all the license header dates.

bobthecow:
Oh, and if you want `@author` attribution, let me know what to use and I'll add that to these files.

Markcial:
I am :+1:  with the license. About the `@author tag`, the details of my name and email so anyone needing something can ping me is fine. `Marc Garcia <markcial@gmail.com>`

bobthecow:
Ok, this has been merged into `develop` and will be in the next release!

Markcial:
Cheers!!! :beers:  congratulations on the new feature!! :smiley:

bobthecow:
This one was all you :)

PR #142, opened 2015-01-27
Shorter syntax for the debug shell

Markcial:
It adds another dependency based on the library SuperClosure, but it leaves a shell call with a shorter syntax. Feel free to modify the function , reject the PR or whatever. I can have this only on my fork, but i thought it could be interesting for you

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
         "symfony/console": "~2.3.10|~2.4.2|~2.5",
         "nikic/php-parser": "~1.0",
         "dnoegel/php-xdg-base-dir": "0.1",
-        "jakub-onderka/php-console-highlighter": "0.3.*"
+        "jakub-onderka/php-console-highlighter": "0.3.*",
+        "jeremeamia/SuperClosure": "@stable"
     },
     "require-dev": {
         "phpunit/phpunit": "~3.7|~4.0",
--- a/src/Psy/functions.php
+++ b/src/Psy/functions.php
@@ -2,6 +2,8 @@
 
 namespace Psy;
 
+use SuperClosure\Analyzer\TokenAnalyzer;
+
 /**
  * Command to return the eval-able code to startup PsySH.
  *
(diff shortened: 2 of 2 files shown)


bobthecow:
I've considered doing things with SuperClosure in the past, but have run into issues where PHP was too dumb about things to let me use it, so I'm not sure I'm comfortable 1) making it a dependency, and 2) recommending it over the `eval` version that works, for what is a relatively small gain :)

```php
eval(\Psy\sh());
\Psy\dbg();
```

Markcial:
its a good point, after 3 hours of looking at the debug_backtrace I had suddenly felt really sad with the PHP functionalities, but i came into this, as a informational note the PR is here, but if you reject it won't be a shame and I will completely understand.

bobthecow:
Now, if you wanted to write a PHP extension that added a `debug` keyword, *that* would be a huge win :)

Markcial:
wish i could, i am not as badass writing extensions as nikic. Maybe in the long run. :)

bobthecow:
Heh. Me too :)

I'm marking this as an enhancement and closing it for now.

bobthecow:
This might actually fix the issues I was running into before: https://github.com/jeremeamia/super_closure/issues/34

… but it's hard to say without extensive testing :)

composer.json
```diff
@@ -17,7 +17,8 @@
         "symfony/console": "~2.3.10|~2.4.2|~2.5",
         "nikic/php-parser": "~1.0",
         "dnoegel/php-xdg-base-dir": "0.1",
-        "jakub-onderka/php-console-highlighter": "0.3.*"
+        "jakub-onderka/php-console-highlighter": "0.3.*",
+        "jeremeamia/SuperClosure": "@stable"
```
GrahamCampbell: This is a super terrible version constraint, and is already broken....

bobthecow: agreed. but we're likely not using it at all, so i didn't worry too much about it :)

Markcial:
Good news everyone, i found a library for php extensions, is possible to have a shorter syntax without troubles only installing the module, some screenshots will follow
captura de pantalla 2015-01-29 a las 23 48 08
captura de pantalla 2015-01-29 a las 23 48 53

The library used is [PHP-CPP](http://www.php-cpp.com/) and it could be easily compiled, but it is a whole new level of dependencies, if you are interested i can open a whole new repository for the module and maybe some helper functions for psysh.

PR #143, opened 2015-01-27
Don't include args in the whereami backtrace.

bobthecow:
Saves us a bit of memory and complexity, at the very least.

We could also probably limit the backtrace to some sane number of stack frames, but that's a bit risky because inside evals or closures or generated code (SuperClosure, etc) it can get deeper than you'd expect.

/cc @blainesch

### DIFF
--- a/src/Psy/Command/WhereamiCommand.php
+++ b/src/Psy/Command/WhereamiCommand.php
@@ -25,7 +25,7 @@ class WhereamiCommand extends Command
 {
     public function __construct()
     {
-        $this->backtrace = debug_backtrace();
+        $this->backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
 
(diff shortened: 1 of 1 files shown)


blainesch:
Didn't realize that was an option :+1:

How much does it actually save though? PHP doesn't copy until you attempt to edit them, do you have any benchmarks?

bobthecow:
Nah. I didn't see any downside, and the manual page on `debug_backtrace` says it's an improvement, so I figured why not? :)

blainesch:
On second though, I might be wrong. I'm assuming it does a "copy on edit" but it probably contains a copy of a non-changed item. For instance, passing a complex object into a method is probably preserved in every state so you can debug it better, thus a copy of each item is potentially generated.

bobthecow:
Cool. We'll ship this change in the next release then :)

PR #145, opened 2015-01-31
Prevent function redeclaration.

pitekantrop:
On a system that has PsySH installed globally, running *any* global composer command on a project that has PsySH installed locally might cause composer to load `Psy\functions.php` from both the global and the local context. This will result in a function redeclaration error.


bobthecow:
Thanks, this has been merged into `develop` and will go out in the next release :)

Issue #99, opened 2014-10-01
Segmentation fault (core dumped)


branquito:
Getting this error when trying to run `psysh` from terminal.

`Segmentation fault (core dumped)`. What might be the problem.

dequis:
Hey! I had the same segfault issue, debugged it a bit and submitted a bug report with patch here: https://bugs.php.net/bug.php?id=69054

More specifically, I got the following ErrorException from laravel:

```
$ php artisan tinker
exception 'ErrorException' with message 'is_file(): open_basedir restriction in effect. File(/etc/xdg/psysh/config.php) is not within the allowed path(s): (/srv/http/:/home/:/tmp/:/usr/share/pear/)' in /home/dx/test/asd/sonar/laravel/vendor/psy/psysh/src/Psy/Configuration.php:132
Stack trace:
#0 [internal function]: Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(2, 'is_file(): open...', '/home/dx/test/a...', 132, Array)
#1 /home/dx/test/asd/sonar/laravel/vendor/psy/psysh/src/Psy/Configuration.php(132): is_file('/etc/xdg/psysh/...')
#2 /home/dx/test/asd/sonar/laravel/vendor/psy/psysh/src/Psy/Configuration.php(105): Psy\Configuration->getConfigFile()
#3 /home/dx/test/asd/sonar/laravel/vendor/psy/psysh/src/Psy/Configuration.php(89): Psy\Configuration->init()
#4 /home/dx/test/asd/sonar/laravel/vendor/psy/psysh/src/Psy/Shell.php(68): Psy\Configuration->__construct()
#5 /home/dx/test/asd/sonar/laravel/vendor/laravel/framework/src/Illuminate/Foundation/Console/TinkerCommand.php(31): Psy\Shell->__construct()
#6 [internal function]: Illuminate\Foundation\Console\TinkerCommand->fire()
#7 /home/dx/test/asd/sonar/laravel/vendor/laravel/framework/src/Illuminate/Container/Container.php(523): call_user_func_array(Array, Array)
#8 /home/dx/test/asd/sonar/laravel/vendor/laravel/framework/src/Illuminate/Console/Command.php(115): Illuminate\Container\Container->call(Array)
[...]

PR #150, opened 2015-02-15
Workaround PHP bug #69054 and other issues with open_basedir

dequis:
[PHP bug #69054][1] is a segfault when calling readline_read_history() with no parameters when open_basedir is set.

This commit also adds a few @ before is_file() calls in Configuration which may throw ErrorException due to open_basedir

Related: #99

[1]: https://bugs.php.net/bug.php?id=69054


bobthecow:
Great! Thank you. This has been merged into `develop` and will be in the next point release.

Issue #153, opened 2015-02-18
code run in a long running operation doesnt show contents after the operation has completed


Markcial:
When you run a code snippet like these : 
<code>
$foo = 5;
while ($foo) {
    echo 'boh!' . PHP_EOL;
    sleep(1);
    $foo--; 
}
</code>

Until the operation has completed the output doesnt return. Will be nice that the output buffering can throw the output on every cycle of operation, it can be done calling the ob_start with implicit flush or calling with max chunksize = 2 so the output flushes every line.

PR #155, opened 2015-02-18
Implicit flush for chunks bigger than one line

Markcial:
solves ticket_135

### DIFF
--- a/src/Psy/ExecutionLoop/Loop.php
+++ b/src/Psy/ExecutionLoop/Loop.php
@@ -61,16 +61,14 @@ class Loop
                     $__psysh__->getInput();
 
                     // evaluate the current code buffer
-                    ob_start();
+                    ob_start(array($__psysh__, 'writeStdout'), 2);
 
                     set_error_handler(array($__psysh__, 'handleError'));
                     $_ = eval($__psysh__->flushCode());
                     restore_error_handler();
 
-                    $__psysh_out__ = ob_get_contents();
                     ob_end_clean();
 
-                    $__psysh__->writeStdout($__psysh_out__);
                     $__psysh__->writeReturnValue($_);
                 } catch (BreakException $_e) {
                     restore_error_handler();


src/Psy/ExecutionLoop/Loop.php
```diff
@@ -61,16 +61,14 @@ public function run(Shell $shell)
                     $__psysh__->getInput();
 
                     // evaluate the current code buffer
-                    ob_start();
+                    ob_start(array($__psysh__, 'writeStdout'), 2);
```
bobthecow: This should be `1` in PHP >= 5.4, since that's no longer a special case.

bobthecow:
My understanding is that the `ob_start` callback is supposed to be for filtering the output buffer, not for executing a side effect. Is using it like this going to cause us any trouble?

Markcial:
It happened to me on some functions when using the shell : https://github.com/bobthecow/psysh/issues/153

bobthecow:
`Shell::writeStdout` always writes full lines, and it appends a newline glyph to the end of input when outputs text that doesn't end in a newline. For example, this:

```php
$foo = 5;
while ($foo) {
  echo '.';
  sleep(1);
  $foo--; 
}
```

... will output this:

```
.⏎
.⏎
.⏎
.⏎
.⏎
=> null
```

Which is certainly not what I'd expect :)

Markcial:
do we have a way to have the feedback on long running operations but without losing the current benefits ? I'd like to have the newlines when i do single line operations, but i would like to have the feedback on expensive operations run in the shell too.

bobthecow:
Yes, I understand the reason for this, and :+1: for that :)

I'm just worried about the fact that we're using a side effect inside the output callback. From reading the documentation, it looks like that callback is only meant to let you modify the text of the output buffer before it's flushed, and it even has warnings about calling functions that could mess with output.

bobthecow:
Got it!

```
>>>  $foo = 5; while ($foo) { echo '.'; sleep(1); $foo--; }
.....⏎
=> null
```

Just a second, I'll push it.

Markcial:
oh nice! great job!

bobthecow:
Here's what I ended up with:

https://github.com/bobthecow/psysh/compare/develop...feature/incremental-flush

It handles both the incremental and the final flush. As a bonus, it now outputs the implicit newline when doing an `ob_end_clean` during error handling, as well!

I'm still a little worried about using a side effect of the callback to actually render our output. With the current default output, I don't think it's a problem, but since it's possible to use other types of output, it could potentially bite us later. That said, it's probably worth letting future us deal with that problem :)

Markcial:
I really appreciate the effort, in our team we use the shell not only for testing but for administrative techie tasks, so this is a really nice feature for us, thanks a lot!

bobthecow:
No problem! This has been merged into `develop` and will go out in the next release.

PR #157, opened 2015-02-19
proposal for #ticket_154 

Markcial:
quirky fix with a cancellable forking loop that aborts long running operations with kill signal without exiting from the psysh shell, can be set as a parameter


Markcial:
This PR might seem unrelated https://github.com/bobthecow/psysh/pull/138 but it has a loop refactor proposal that could involve this forking loop, linked here just as a side not for any visitor in this thread.

bobthecow:
I haven't taken any time to dig into why, but it's definitely not working right just yet:

screen shot 2015-02-19 at 11 16 41 pm

Markcial:
Ok, thanks for the hint, don't waste your time on that. I will check for that later and write a test covering the case.

bobthecow:
Do you mind re-opening this pull request against develop?

Markcial:
done, i will close this PR then

Issue #154, opened 2015-02-18
Possibility to cancel long running operations with Ctrl-C


Markcial:
When calling long time running operations a nice feature could be the chance to cancel that operation without exiting from the shell, it can be done with Ctrl-C or any signal that will tell the shell it have to kill the spawned worker but not the main thread, for then creating a new worker with the state from the main worker, like the savegame function but with rollback.

bobthecow:
:+1: Yes, let's :)

Markcial:
Experiment in development here => https://github.com/bobthecow/psysh/pull/163

eode:
..would this have the happy consequence of dropping back to the prompt instead of quitting, or should I open a new ticket for that?

sehgal10:
So whats the progress or alternative to this? Do we have any workaround than Ctrl + C?

bobthecow:
Not right now, but pull requests are welcome :)

bobthecow:
This is available in the latest release!

https://github.com/bobthecow/psysh/releases/tag/v0.12.13

onlime:
Just for Laravel folks (or other weirdos) who stumble accross this and have used some kind of Bash alias trick to loop over a command like `php artisan tinker`...

```bash
alias patt="while true; do php artisan tinker; done"
```

[PsySH v0.12.13](https://github.com/bobthecow/psysh/releases/tag/v0.12.13)'s new SIGINT handling broke this, so you can no longer exit the loop with `Ctrl-C`. Here's my workaround:

```bash
alias patt='trap "kill 0" SIGINT; while true; do php artisan tinker || break; done; trap - SIGINT'
```

Like this, you can still restart Tinker with `Ctrl-D` and exit it with `Ctrl-C`.

bobthecow:
@onlime that's an interesting workflow! i wonder if there's a way we can support it more directly, rather than relying on hacks?

interestingly, your first loop exits for me if i double-tap `ctrl-c`.  it has to be pretty quick, but it works the same as with previous psysh versions.

PR #163, opened 2015-02-24
Cancelable forking loop

Markcial:
cancellable forking loop

See #154

### DIFF
--- a/src/Psy/Configuration.php
+++ b/src/Psy/Configuration.php
@@ -12,6 +12,7 @@
 namespace Psy;
 
 use Psy\Exception\RuntimeException;
+use Psy\ExecutionLoop\CancellableForkingLoop;
 use Psy\ExecutionLoop\ForkingLoop;
 use Psy\ExecutionLoop\Loop;
 use Psy\Output\OutputPager;
@@ -33,7 +34,7 @@ class Configuration
         'defaultIncludes', 'useReadline', 'usePcntl', 'codeCleaner', 'pager',
         'loop', 'configDir', 'dataDir', 'runtimeDir', 'manualDbFile',
         'requireSemicolons', 'historySize', 'eraseDuplicates', 'tabCompletion',
-        'tabCompletionMatchers',
+        'tabCompletionMatchers', 'cancellable',
     );
 
     private $defaultIncludes;
@@ -49,6 +50,7 @@ class Configuration
     private $useReadline;
     private $hasPcntl;
     private $usePcntl;
+    private $cancellable;
     private $newCommands = array();
     private $requireSemicolons = false;
     private $tabCompletion;
@@ -826,7 +828,11 @@ class Configuration
     {
         if (!isset($this->loop)) {
             if ($this->usePcntl()) {
-                $this->loop = new ForkingLoop($this);
+                if ($this->getCancellable()) {
+                    $this->loop = new CancellableForkingLoop($this);
+                } else {
+                    $this->loop = new ForkingLoop($this);
+                }
             } else {
                 $this->loop = new Loop($this);
             }
@@ -870,6 +876,22 @@ class Configuration
         if (isset($this->shell)) {
             $this->shell->addTabCompletionMatchers($this->tabCompletionMatchers);
         }
+    }
+
+    /**
+     * @param $bool
+     */
+    public function setCancellable($bool)
+    {
+        $this->cancellable = $bool;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getCancellable()
+    {
+        return $this->cancellable;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


Markcial:
With this refactor the forking loop keeps variables in scope without sending serialized variables over the pipes, seen here => https://github.com/bobthecow/psysh/issues/106.

The idea with this forking loop is to keep the main process open and spawn new ones that die afterwards, instead of spawning a new child, passing the vars, kill the parent etcetera.

bobthecow:
I'm about to ship a patch release without your changes in it. I'm not ignoring you, I'm just ignoring your pull requests until I have time to actually review ;)

Markcial:
don't worry, i just do the PR so the latest versions can be available for anyone interested. I know you are busy, take it easy, it's ok really. I'd rather prefer you to do the reviews without the hustle, but thanks for pinging me back about this :D

Markcial:
The support for php5.4 has been dropped? why the style CI is asking me to replace the array function calls by the square brackets?

bobthecow:
Because you need to merge `develop` (or rebase) to get the `.styleci.yml` file.

Markcial:
ah ok, just a moment then, thanks

bobthecow:
(I just updated your title to what the fix does, and your first comment to link this PR with the issue it fixes)

Markcial:
State of this PR, does it need to cover some more cases? I will write the tests for that. Currently this ForkingLoop does not uses pcntl at all, only uses a single thread and preserves state by catching exceptions and calling again the run method, if you are not confident enough with the pcntless forking loop, i might be able to split it apart to another forking loop, having both of them, the pcntl one and this one, letting the config serve the pcntl by default.

I guess that having this on master would be nice the public api from the shell opens a lot of possibilities, Expl.: remote debugging with a server querying the public execute method on a cli server with the shell.

If you need something that i am able to do, just tell me. Thanks

bobthecow:
I think it'd be nice to have on master as well. Being able to execute arbitrary things on the execution loop enables a lot of things: the timer command, upgrading commands with a target to be able to reason about return values and not just variables (like `dump $foo->bar()`), etc.

We're going to need the forking, unfortunately. PHP just dies far too often to run without it, and there's really no way for us to stop it.

bobthecow:
Oh, I just thought of a really simple "cancelable loop" … I'm not sure if it's the best way to do it, but it'd probably work: use the normal pcntl-based forking loop, and register a signal handler that listens for ctrl+c and sends the current child fork a sigkill. The parent would "recover", effectively canceling the current operation. Thoughts on that?

Markcial:
I don't mind splitting the forking loop on two, the cancel can be done on both of them. But when i said if there were some more cases to cover i meant the ones when php dies without way to recover. This way i can add some more tests. And with that maybe some day PHP might not need to fork processes continously and it might be able to deal with the errors gracefully without shutting down then the pcntl will not be necessary.

The pcntl loop has its owns pros and cons, it can be used without worrying about it to have a fatal error that closes everything but the closures or other non serializable objects are missing when transferring information to the child thread.

The loop that doesn't uses pcntl can preserve the context of variables and non serializable objects, but on non recoverable fatal errors it exits and doesn't recovers.

Based on that, my vote is to have both of them and the user chooses what he wants, preserve callbacks but without fatal error recovery, or miss the non serializable objects but a rock heavy shell.

So if you agree with that, i will split on two forking loops with their own cancellable extends.

bobthecow:
There already are two loops. The forking one uses pcntl. The other loop doesn't. The user can already choose between them (by disabling pcntl in their config). I'm not sure what you're suggesting doing differently :)

Markcial:
Yes, there are two loops, but in this PR i did a complete rewrite of both of them.

For example, the forking loop i did doesn't uses the pcntl no more:

```php
    protected function replay(Shell $shell, $executionClosure)
    {
        try {
            // get input
            $shell->getInput();

            $shell->beforeLoop();
            $executionClosure($shell);
            $shell->afterLoop();
        } catch (BreakException $e) {
            throw $e;
        } catch (\Exception $e) {
            $shell->resetCodeBuffer();
            $shell->writeException($e);
            $this->replay($shell, $executionClosure);
        }
    }
```

With this, whenever the exception is successfully handled the shell goes on with the same context calling the same function (replay) again, even, closures or database connections, file resources, etc, but when the php throws a fatal non recoverable error, the loop just stops, the flaw that the pcntl doesn't have. But if you want to use a lot of anonymous functions, file resources or database connections, the pcntl choice brings a little trouble there.

bobthecow:
But that's the "forking" part of "forking loop" :)

Markcial:
Yes, maybe it is for the name that it is confusing. I did a "recursive" loop

GrahamCampbell:
Any news on this?

neronmoon:
+1 here

castarco:
If this branch is dead... maybe it should be removed.

PR #169, opened 2015-03-01
Added ".styleci.yml" And Applied CS Fixes

GrahamCampbell:


composer.json
```diff
@@ -20,10 +20,7 @@
         "jakub-onderka/php-console-highlighter": "0.3.*"
     },
     "require-dev": {
-        "phpunit/phpunit": "~3.7|~4.0",
-        "symfony/finder": "~2.1",
```
bobthecow: `symfony/finder` is required by `bin/compile`

composer.json
```diff
@@ -20,10 +20,7 @@
...
-        "squizlabs/php_codesniffer": "~2.0",
-        "fabpot/php-cs-fixer": "~1.3"
```
bobthecow: By removing these from composer dev requirements, you force contributors to maintain a global installed copy that stays in sync with whatever version we expect to have. Is there any downside to leaving them here so that people can still run `vendor/bin/php-cs-fixer fix` before committing?

composer.json
GrahamCampbell: @bobthecow phpunit requires symfony finder. I can add it back if you like. :)

composer.json
GrahamCampbell: > By removing these from composer dev requirements, you force contributors to maintain a global installed copy that stays in sync with whatever version we expect to have. Is there any downside to leaving them here so that people can still run vendor/bin/php-cs-fixer fix before committing?

I think it's fine for people to have them globally installed otherwise we're adding time to our travis build. Some projects even expect contributors to have a global install of phpunit too.

GrahamCampbell: StyleCI makes it super easy for contributors to download a diff then can apply using "git apply" anyway. I think this is arguably faster than running php-cs-fixer on their system - it runs much slower for people with xdebug installed for example.

bobthecow: But that workflow requires people to commit things that are wrong. It doesn't empower them to actually do it right the first time. They essentially have to break a build to see that they've done something wrong, and that's not a good experience at all. Travis isn't there as a first defense, it's the backup in case a failing test gets through. StyleCI should be the same.

GrahamCampbell: StyleCI does have a new feature coming up where by you will be able to run a command line tool that will upload the code to our servers, analyse it, and then download and apply the fixes.

GrahamCampbell: > But that workflow requires people to commit things that are wrong.

People can still use global installs. I'd totally recommend that. php-cs-fixer works well as a global install.

GrahamCampbell: > It doesn't empower them to actually do it right the first time.

I normally write code that passes StyleCI's checks first time, lol.

bobthecow: Me too — save for when new rules are introduced between builds, see FriendsOfPHP/PHP-CS-Fixer#1038 — but most people don't.

It feels like that's moving in the wrong direction. It's fine for PHPUnit, because it's old enough to be almost completely stable and backwards compatible. php-cs-fixer is such a moving target that just using whatever version someone happens to have globally installed is _almost guaranteed_ to be wrong.

GrahamCampbell: Is there a way to skip install on travis?

GrahamCampbell: Without having 2 different composer.json files?

GrahamCampbell:
@bobthecow I've restored stuff to the composer.json as you asked. :)

GrahamCampbell:
I still worry that they're going to cause conflicts on global installs, say if people want to use php-cs-fixer 2.0, or an older version of phpcs.

bobthecow:
Yeah. Global installs in Composer are pretty broken. I feel like `composer g update` and `composer g require` should imply `--no-dev` and `--update-no-dev` unless specified otherwise. TBH, I'm considering removing global install instructions in favor of just using the phar if you want a global.

composer.json
bobthecow: Yeah, but since I'm using it, I'd prefer not to rely on a dependency of a dependency, since that's not guaranteed to remain the case.

composer.json
bobthecow: Good question. We could rewrite the requirements before building :P

bobthecow:
I'll merge this because it's closer to correct. You're right, that we'll need to figure out global installs, but this isn't any worse with global installs than it already is :)

Issue #166, opened 2015-02-26
Plugin perspective


Taluu:
What would be awesome would be, using Psysh, to be able to register extensions, so that it may register custom commands (without going through a PR here), outputs, presenters, ... or whatever else.

Just launching the discussion here, so that we may maybe get some ideas. I was thinking of something like [how Twig does it](https://github.com/twigphp/Twig/blob/master/lib/Twig/ExtensionInterface.php) (Or [Symfony2 and its bundles](https://github.com/symfony/symfony-standard/blob/2.7/app/AppKernel.php#L8), as this is more complete but also [more tedious](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpKernel/Kernel.php#L455))

Any thoughts ?

Markcial:
You can roll your own commands with the configuration, http://psysh.org/#configure you only need to have the psr0 path defined, extend the basecommand class, and then create the new instance in the commands parameter. Anyways, to load commands from a subfolder if they extend  a class might be another option.

Taluu:
Yep, but with a Plugin, you may also register not only commands, and I think that this is baseline for such a system. Anyhow, it seems @bobthecow that is interested in such a discussion, as it seems he already tried to approach such a system, without being satisfied.

Just creating the issue so that we can brainstorm on that matter.  ;)

It is already great though that we can register commands... Maybe push the config thing but for plugins ? So that they may live as full fledged plugins.

Markcial:
The ways that came into my mind of registering plugins are; with a base folder that has classes that implements some kind of `PluginInterface`, with a command with calls like `plugin add "class"` or with the configuration. My two cents here.

Taluu:
I don't know if registering a plugin through the app (via a `plugin` command) wouldn't be too late, as the app is already launched. I was also thinking of an interface to implement, and then somewhere (such as the config file ?) register the plugin.

when the app is launched, it then checks all the plugin to instanciate, and registers all the stuff that goes along with it (such as commands, printers, presenters, ....)

Markcial:
Well then it depends which layers are explosed publicly, @bobthecow might have a better knowdledge about this. If something cames up i might be able to help a little bit, i will keep an eye open o this thread then.

bobthecow:
Here's a dump of what I'm thinking:

[...]

PR #171, opened 2015-03-02
plugin manager

Markcial:
What it currenlty does:
- successfully creates a configuration array after registering the plugins.
- add matchers, commands or presenters to the configuration.
- give static access to the configuration or plugins.
  Example: 
  <pre>
  \Psy\Plugin\DemoPlugin::register();
  \Psy\Plugin\Manager::getConfiguration(); // will return the democommand added in the configuration
  </pre>
  What it is not yet able to do.
- Magically discover plugins by vendor path.
- Replace properties in the configuration keeping the newest ones.


src/Psy/Plugin/AbstractPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * @return string
+     */
+    final public static function getName()
+    {
+        $class = new \ReflectionClass(get_called_class());
+
+        return preg_replace('#Plugin$#', '', $class->getShortName());
```
Taluu: I'm not sure that this is really relevant, I think it should be an abstract

src/Psy/Plugin/Manager.php
```diff
@@ -0,0 +1,32 @@
...
+    protected static $plugins = array();
+
+    /**
+     * @param AbstractPlugin $plugin
+     * @param $name
+     */
+    public static function register(AbstractPlugin $plugin, $name)
+    {
+        self::$plugins[$name] = $plugin;
+    }
```
Taluu: I'm not quite sure that this class should be static (Or otherwise it should be called something else, such as `PluginRegistry` ?) ; thus, we would need to pass the `Manager` instance to the `register` method of the Plugin

src/Psy/Plugin/Manager.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * @param array $configuration
+     *
+     * @return array
+     */
+    public static function getConfiguration($configuration = array())
+    {
+        foreach (self::$plugins as $plugin) {
+            $configuration = $plugin::getConfiguration($configuration);
```
Taluu: you forgot some things here. You're overwrting the `$configuration` with the last plugin found in the `$plugins` array, and overwriting the parameter `$configuration`, so it is not correct anymore with the 2nd+ plugin...

src/Psy/Plugin/Manager.php
Markcial: What is the reason behind that? why the plugin has to be able to modify the manager? The manager should not be accessible, if the manager goes public we might be needing some intermediate layer for protection like the registry one you said before.

src/Psy/Plugin/AbstractPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    public static function register()
+    {
+        Manager::register(new static(), static::getName());
+    }
+
+    /**
+     * @return string
+     */
+    final public static function getName()
```
Taluu: only the `register` function should be static imho, as it initializes the plugin

src/Psy/Plugin/AbstractPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+     * @return array
+     */
+    final public static function getConfiguration($configuration = array())
+    {
+        return array_merge_recursive(
+            $configuration,
+            array(
+               'commands'   => static::getCommands(),
+               'presenters' => static::getPresenters(),
+               'matchers'   => static::getMatchers(),
```
Taluu: all the getX methods should be protected imho, and this `getConfig` should be the only one accessible with the `getName`

src/Psy/Plugin/AbstractPlugin.php
Markcial: I am not sure too, maybe automation is not the solution, but maybe writting the name for all the plugins is not necessary. I leave this like this temporarily, but i am not convinced too.

src/Psy/Plugin/Manager.php
Taluu: With the method you used, it is accessible though... it should also check that the plugin is not already registered

src/Psy/Plugin/Manager.php
Markcial: the configuration is merged, the configurations are preserved, but the commands, matchers or presenters are not removed.

Taluu: oh, them I don't think it should merge anything, it should just return the brute configuration rather than try to merge it with something else

Even though it is really subjective, I think it could be confusing (at least I was confused :x)

src/Psy/Plugin/AbstractPlugin.php
Markcial: what if they only want the matchers for the plugin but not the presenters or the commands? I do not see a problem to expose publicly that methods, are only getters with static data.

src/Psy/Plugin/AbstractPlugin.php
bobthecow: tbh, how much does the plugin name matter?

src/Psy/Plugin/AbstractPlugin.php
bobthecow: what use would you have for an instance of a plugin?

src/Psy/Plugin/AbstractPlugin.php
Taluu: Almost none but still, for a registry (such as the `Manager`) it could be good to ensure that a plugin is registered only once. Never tried to register a same command more than once, but wouldn't it cause problems ?

bobthecow: I think it'd just override the previous one.

src/Psy/Plugin/AbstractPlugin.php
Taluu: as an object point of view I guess (I still have Twig's extensions system in mind, which allows from an object pov to instantiate an extension)

src/Psy/Plugin/AbstractPlugin.php
bobthecow: It'd be super useful for plugins which depend on plugins. For example, Drupal, Laravel and EZ all depend on some parts of Symfony. So if there were a Symfony plugin, perhaps those others would want the Symfony commands, but not presenters or matchers. Or if there were a Doctrine plugin, I could see wanting the ability to present and match Doctrine models, but not expose all the Doctrine console commands, because they are legion.

src/Psy/Plugin/AbstractPlugin.php
bobthecow: But it doesn't make sense as an object. It's an object with no responsibility :)

src/Psy/Plugin/AbstractPlugin.php
Taluu: Seems fair, even though (still on a (quite) subjective object pov) I don't know if there is really any interest for the thing that registers the plugin to have any access to other methods

src/Psy/Plugin/AbstractPlugin.php
Taluu: Yes, but it is a value-objet, not just a class :D

bobthecow: Honestly, if PHP could autoload functions, I'd prefer it be a function. [There's absolutely no reason to be beholden to objects](http://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html) :)

Taluu: Then @Markcial's register method in the `AbstractPlugin` is wrong. Why instanciate the plugin when all it should do is just a  `__CLASS__` instead ?

(Still subjective, but I really don't like static if I can help otherwise :D)

Markcial: `new __CLASS__` is an alias of `new static`, so it's only about coding style.

Taluu: I was refering to only register the class name, not make an instance of it : `Manager::register(__CLASS__, static::getName())`

Taluu: Mmh nay it is wrong, as `__CLASS__` is not statictly bound, it will return "AbstractPlugin" here. So it should use `get_called_class()` instead

bobthecow:
One more piece, configuration-wise. I think we should have a `disableAutoRegister` (or some other name meaning something like that) static method on the plugin manager which could be called in a config file or shell wrapper to prevent automatically registering plugins.

Markcial:
Good catch, after the `$this->init` which loads the files and stuff might be ok, right?

bobthecow:
Hmm. The order things happen might need to be tweaked. We want presenters defined in the config file to be registered after automatically discovered ones, so that they can be locally overridden, but we also want to be able to suppress automatic registration inside the config file.

Markcial:
ok,then i temporarily add the function call at the end, later we could be able to add test coverage or tweak it as needed.

src/Psy/Configuration.php
```diff
@@ -96,6 +98,27 @@ public function __construct(array $config = array())
...
+            $this->loadConfig(Manager::getConfiguration($config));
+        }
+    }
+
+    /**
+     * @param $bool
+     */
+    public function setRegisterPlugins($bool)
+    {
+        $this->registerPlugins = $bool;
```
bobthecow: Please add `(bool)` here.

Markcial: :+1:

src/Psy/Plugin/AbstractPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+
+        return preg_replace('#Plugin$#', '', $class->getShortName());
+    }
+
+    /**
+     * @param array $configuration
+     *
+     * @return array
+     */
+    final public static function getConfiguration($configuration = array())
```
bobthecow: This method doesn't feel quite right. It doesn't feel like it should be this plugin's responsibility to worry about any other configuration, and in the current implementation, every plugin has a chance to mess with the global configuration. Yeah, I get that this is `final`, but it still feels wrong. The manager should be responsible for doing the merging.

Markcial: totally agree, i am not convinced yet, i was waiting to get with a good idea to put there, let's see.

Taluu: I've been thinking about this method, and I do agree that having "2" entry point (one through a `getConfiguration` and one through the getters) is two much. More over, this method is doing a merge to "chain" the methods, and I don't think that this should be the scope of this method. it should just return a plain array if it is left as-is

Taluu: do basically, :+1: @bobthecow

bobthecow: Why not have the manager call `getCommands`, `getMatchers`, etc on each plugin and merge those, respectively? I'd much prefer something explicit like that at the manager level.

Taluu: yes that's what I was referring to, let the plugin manager call by itself the getters

src/Psy/Plugin/AbstractPlugin.php
Markcial: yes, the name is the unique key indexerv see  src/Psy/Plugin/Manager.php line 16, so if anyone adds a plugin with the same name it will be overwritten like a FIFO pile.

src/Psy/Plugin/Manager.php
Markcial: the methods indeed are accessible, but no the plugins variable. maybe an exception if someone tries to register an already registered plugin, but why another intermediate layer for plugin handling?

src/Psy/Plugin/AbstractPlugin.php
Markcial: objects can hold static calls too, it doesn't matter, you can call `ClassName::method` or `$object::method`

src/Psy/Plugin/AbstractPlugin.php
bobthecow: But what if there are two people with `DevToolsPlugin` or `AwesomePlugin`? Is disambiguating by the first part of the short class name the best way to go?

Markcial: So true, at first was thinking about using the full namespaced class. Anyways i am not yet convinced about this, maybe is hould open up this method and get a default naming.

bobthecow: Add an abstract static function that plugins have to implement and call it a day. That way plugins can just use their Packagist package names. Mine can be called `bobthecow/psysh-devtools` and yours can be called `markcial/psysh-devtools`, and we'll never collide :)

src/Psy/Plugin/Manager.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Psy\Plugin;
+
+class Manager
```
bobthecow: For consistency with existing managers, this should prolly be called `PluginManager`.

Markcial: ok

bobthecow:
More commentary back on the main thread :)

https://github.com/bobthecow/psysh/issues/166#issuecomment-77422344

src/Psy/Plugin/PluginManager.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+namespace Psy\Plugin;
+
+class PluginManager
```
GrahamCampbell: Should this class be made final perhapse?

Issue #170, opened 2015-03-01
Let's sort out global install recommendations.


bobthecow:
`composer global require` is sub-optimal, because it puts psysh and all its dev and non-dev dependencies in a pile with anything else that's globally installed. It turns out that this approach doesn't scale. So maybe we shouldn't do that.

What should we do instead?

See #169

GrahamCampbell:
Maybe we should recommend using a phar file, and adding it to your PATH in windows, or the equivalent for other OSs.

Markcial:
Hi, maybe an import function encapsed on the Psysh namespace might help, like the ones encapsed on the `functions.php` file.

in pseudocode might look like this :
```php
namespace Psy;

function import($library) // library vendor Ex: "psy/psysh"
{
   static $loader;
   if (isset($loader)) {
     $loader = new Loader();
   }
   // when looking for library, check for its composer file with the autoload data, if library found but without composer warn about this.
   // look for local dependencies in vendor folder, return on success
   // look for global dependencies in the home folder, return on success
   // look for phar dependencies finally, return on success.
   // throw warning if libraries not found.
}
```
for the loader you might use one encapsed in the psysh namespace with the full functionalities, you can check out [this one](https://github.com/Markcial/stego/blob/master/src/Stego/Loader.php), feel fre to use it changing the name. It's ok for me.

This way, you can do a file like this one :
```php

require 'psysh.library';

use Psy;
import('symfony/console');

$console = new \Symfony\Component\Console\Application();
```
or
```php
require 'psysh.library';

Psy\import('symfony/console');

$console = new \Symfony\Component\Console\Application();
```

bobthecow:
This was fixed with the "bin launcher" support in v0.5.0. Now, you can install globally and locally, and it will run the version you have installed locally, even if you just do `psysh` to call the global phar or composer installation.

PR #172, opened 2015-03-05
Make psysh play nicer as a global install

bobthecow:
Use a globally installed `psysh` bin script or phar as a "launcher" for a local version, if possible. Starting from the current working directory and moving up, check for 'psy/psysh' as a composer project or dependency. If found, load that one instead of the globally installed version.

This is basically what Gulp and a bunch of other JS projects do with their global installs as well.

This is a partial fix for #170. I think the rest of the fix is to stop recommending the non-phar install at all, in anything but a project scope. Instead, recommend installing the `psysh` phar somewhere in `$PATH`.

### DIFF
--- a/bin/psysh
+++ b/bin/psysh
@@ -82,9 +82,8 @@ if (!class_exists('\Psy\Shell')) {
     } elseif (is_file(__DIR__ . '/../../../autoload.php')) {
         require __DIR__ . '/../../../autoload.php';
     } else {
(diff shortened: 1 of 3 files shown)


bobthecow:
@GrahamCampbell Thoughts on this approach? I feel like this way we can recommend the global install, but still be able to play nice with users who have, e.g., a local project which depends on older psysh. They could still just run `psysh` from their project directory and it'd work.

PR #174, opened 2015-03-06
Re-added branch alias

GrahamCampbell:


GrahamCampbell:
I'm confused as to why you have a develop branch tbh. It would make more sense to commit anything going into the next patch release straight to the master, and use develop to work on the next minor release.

bobthecow:
There is a development branch because of http://nvie.com/posts/a-successful-git-branching-model/

GrahamCampbell:
I don't think that makes sense. We have tags that do the same thing...

GrahamCampbell:
No major framework follows that model. Work on patch releases never goes in a develop branch for any decent framework I've ever come across.

bobthecow:
You're describing git-flow. He calls 'em "hotfixes", you call em "work on patch releases" but it's the same basic model. Two development paths, one for small things that should be applied directly to master, the other for bigger changes, possibly backwards compatibility breaks, which should go in a development branch.

GrahamCampbell:
Right, you have been working on all the 0.3.x and 0.4.x patch releases on the develop branch, rather than on the master?

bobthecow:
To be honest, developing a framework is one thing. But for a library the size of this, there is seldom need for two parallel development tracks. When that need arises, for example to revert a backwards compatibility break in a stable release, I do a patch off of master. But for the most part, non-urgent small changes get dropped into develop, and big changes get merged into develop as they're ready, and I try to release fairly soon after things get merged. If there are only small changes waiting to ship, it's a patch release. If there are big changes too, I ship a point release. I don't see much need for more structure than that, for something the size of a project like this.

Given that, and how often we end up needing "hotfix" type patches, it wouldn't _hurt_ to only have one master/develop branch, and to occasionally just deal with the awkwardness of releasing a patch by starting a branch at the last tag. But ultimately, either way works and I haven't spent much time thinking about it :)

Issue #175, opened 2015-03-09
require a non existant file throws a uncatchable exception


Markcial:
when trying to require a non existant file, the php throws a E_COMPILE error that cannot be intercepted by the error handler on the psysh shell, maybe a code cleaner pass might help, but it needs to check all include paths and resolve to a file, check if exists, and expand the node if it is a variable, callable, function, etc.

I will try to help with this, but i am not very familiar with code cleaner.

PR #176, opened 2015-03-09
Fix uncatchable exception when requiring non-existent file

Markcial:
For the require, and require_once the code cleaner wraps the nodes inside this :

``` php
return call_user_func(function ($class) {
  if (stream_resolve_include_path($class) {
    return require $class;
  } else {
     trigger_error(sprintf('require(%s): failed to open stream: No such file or directory in php shell code on line %d', $class, $node->getLine()));
 }
}, $class );
```

If the error is not enough, an exception can be added instead. but i thought that the warning might be enough because is the first thing that php throws on this kind of error.

See #175

### DIFF
--- a/src/Psy/CodeCleaner.php
+++ b/src/Psy/CodeCleaner.php
@@ -26,6 +26,7 @@ use Psy\CodeCleaner\LeavePsyshAlonePass;
 use Psy\CodeCleaner\LegacyEmptyPass;
 use Psy\CodeCleaner\MagicConstantsPass;
 use Psy\CodeCleaner\NamespacePass;
+use Psy\CodeCleaner\RequirePass;
 use Psy\CodeCleaner\StaticConstructorPass;
 use Psy\CodeCleaner\UseStatementPass;
 use Psy\CodeCleaner\ValidClassNamePass;
@@ -86,6 +87,7 @@ class CodeCleaner
             new ValidClassNamePass(),
             new ValidConstantPass(),
             new MagicConstantsPass(),
+            new RequirePass(),
         );
     }
 
--- a/src/Psy/CodeCleaner/RequirePass.php
+++ b/src/Psy/CodeCleaner/RequirePass.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Psy\CodeCleaner;
+
+use PhpParser\Node;
+
+class RequirePass extends CodeCleanerPass
+{
+    public function enterNode(Node $node)
+    {
+        if ((
+            $node instanceof Node\Expr\Include_ &&
+            in_array($node->type, array(3, 4)) &&
+            !$node->getAttribute('included')
+        )) {
+            $injected = clone $node;
+            $injected->setAttribute('included', true);
+
+            return new Node\Expr\FuncCall(
+                new Node\Name('call_user_func'),
+                array(
+                    new Node\Arg(
+                        new Node\Expr\Closure(
+                            array(
+                                'params' => array(new Node\Param('class')),
+                                'stmts' => array(
+                                    new Node\Stmt\If_(
+                                        new Node\Expr\FuncCall(
+                                            new Node\Name('stream_resolve_include_path'),
+                                            array(new Node\Arg(new Node\Expr\Variable('class')))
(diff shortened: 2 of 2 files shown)


bobthecow:
Thanks for taking a shot at this.

All things being equal, I prefer code cleaners that don't actually change the user's code. If there's a way to enforce or warn or throw exceptions from the cleaner itself rather than replacing a block of input with completely different code, we should do that :)

As far as the result, we should make it match exactly what PHP does, but without failing. So if PHP has a fatal error, we should throw a FatalErrorException. If PHP just triggers a warning, we can trigger a warning. Our messages and codes and everything should be as close to the PHP native functionality as possible.

Ideally, the user will never even know we did anything. That's what we're going for here.

bobthecow:
(I just updated your title to what the fix does, and your first comment to link this PR with the issue it fixes)

Markcial:
Ok, sorry, i will be more descriptive on the PR. from now on. About this, you are right, but it's a bit complicated with the throw keyword, it is a little difficult to use.

I tried with :
```php
php > return (stream_resolve_include_path($file) && require $file) || (throw new Exception);
PHP Parse error:  syntax error, unexpected 'throw' (T_THROW) in php shell code on line 1

Parse error: syntax error, unexpected 'throw' (T_THROW) in php shell code on line 1
```
or
```php
php > return (stream_resolve_include_path($file) && require $file) ?: (throw new Exception);
PHP Parse error:  syntax error, unexpected 'throw' (T_THROW) in php shell code on line 1

Parse error: syntax error, unexpected 'throw' (T_THROW) in php shell code on line 1
```
then i thought, maybe trigger_error.. but then:
```php
php > return (stream_resolve_include_path($file) && require $file) ?: trigger_error('error');
PHP Fatal error:  Call to undefined function  trigger_error() in php shell code on line 1

Fatal error: Call to undefined function  trigger_error() in php shell code on line 1
````
and
```php
php > return (stream_resolve_include_path($file) && require $file) || trigger_error('error');
PHP Fatal error:  Call to undefined function  trigger_error() in php shell code on line 1

Fatal error: Call to undefined function  trigger_error() in php shell code on line 1
```

I am un happy too with the solution, but my PR is here if anyone needs it, or it could help to have anyone a better idea.

Anyways, my solution has a serious flaw, the require has lost all the context from the vars, so maybe needs to be really heavily rethinked, but whatever, i will do a newer commit if i discover how to improve it.

bobthecow:
My priority is always to execute code that's as close as the user entered as possible. This is why ValidClassName and other passes like that just validate what they can, and let everything else through. It's better to let a user fail because they did something dumb than to do something unexpected to their code trying to save them from themselves. We shouldn't inconvenience people doing the right thing in the name of protecting people that are trying to do the wrong thing :)

So far, we've just been letting things through if they're variables or expressions. See, for example, every validation in ValidClassNamePass:

https://github.com/bobthecow/psysh/blob/master/src/Psy/CodeCleaner/ValidClassNamePass.php#L115

I don't know that there is a better solution, as most expressions could have side effects, so we don't want to just evaluate them and see what they are. Variables should be okay, but array or property access wouldn't be.

We might not need a better solution, to be honest. I'm hopeful that more and more things will be catchable as PHP continues to improve. See, for example, the [Catchable "call to a member function of a non-object" Catchable "call to a member function of a non-object"](https://wiki.php.net/rfc/catchable-call-to-member-of-non-object) RFC in PHP 7. A few more improvements like that, and we can just let the language do its job and catch the errors and exceptions that result :)

Markcial:
I know that newer versions of PHP deal more gracefully with stuff like that, glad to know they are into it. If we are not able to evaluate nothing, i am completely stuck on this. If some day i can work a solution that makes not necessary to evaluate the content sent by the user i will post it here. Thanks

bobthecow:
You know, it looks like PHP isn't getting any better at handling this case. Screw it. Let's do runtime validation for things like this when we can't do static analysis.

I'm cherry-picking this into the `develop` branch.

bobthecow:
This is now in `develop` and will go out in the next stable release. Thank you for contributing :)

PR #181, opened 2015-03-19
[Feature]multi line string editing

Volter9:
Hello there!

I recently discovered this awesome REPL for PHP (yesterday, actually :smiley: ). PsySH is great, but there's only one little detail that I miss from PHP's interactive mode (aka `php -a`), it's multiline string editing:

Multiline string edit

That's why I decided to "fix" the situation, behold of `string multi line edit` :smile::

Working example

And current version with the same code:

Current version non working example

Also, I decided to extract `$codeBuffer` and `codeBufferOpen` in separate class. I hope you okay with that, and also added @author tag in this class, I hope you're okay with that too (otherwise feel free to delete it) :blush: 

Thanks for PsySH! :smiley: 

P.S.: I didn't wrote tests yet for `\Psy\CodeBuffer`, do you want me to write unit tests for it, or it's fine?
P.S.S.: Also not sure about DI in `\Psy\Shell`, is it supposed to be injected through Configuration class and public getter or via Shell's setter?

### DIFF
--- a/src/Psy/CodeBuffer.php
+++ b/src/Psy/CodeBuffer.php
@@ -27,6 +27,7 @@ class CodeBuffer
     private $buffer;
     private $open;
 
+    private $calculatedDepth;
     private $depth;
     private $inverse;
 
@@ -62,7 +63,7 @@ class CodeBuffer
      */
     public function isOpen()
     {
-        return $this->open && $this->depth > 0;
+        return $this->calculatedDepth > 0;
     }
 
     /**
@@ -120,17 +121,13 @@ class CodeBuffer
 
         if ($lastCharacter === $this->openOperator) {
             $this->open = true;
+
             $code = substr(rtrim($code), 0, -1);
-        } elseif ($this->depth >= 0) {
-            $this->open = true;
-        } elseif ($this->depth <= 0) {
+        } else {
             $this->open = false;
         }
 
-        if ($lastCharacter === ';') {
-            $this->open = false;
-            $this->depth = 0;
-        }
+        $this->calculatedDepth = $this->depth + $this->open;
 
         $this->buffer[] = $code;
     }
@@ -151,5 +148,6 @@ class CodeBuffer
         $this->buffer = array();
         $this->open = false;
         $this->depth = 0;
+        $this->calculatedDepth = 0;
     }
 }
--- a/src/Psy/CodeBuffer.php
+++ b/src/Psy/CodeBuffer.php
@@ -1,0 +1,155 @@
+<?php
+
+/*
+ * This file is part of Psy Shell
+ *
+ * (c) 2012-2014 Justin Hileman
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy;
+
+/**
+ * Code buffer, keeps track of inputed code.
+ *
+ * @author volter9
+ */
+class CodeBuffer
+{
+    // Characters
+    private $openOperator = '\\';
+    private $oneLevel  = '/(?<!\\\\)(\'|")/';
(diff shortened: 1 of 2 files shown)


bobthecow:
Thanks for trying PsySH, and for reporting the issue, and thanks for taking a shot at this! I never noticed that single quoted multiline strings were breaking.

The trouble with regexp-based parsing is that it doesn't work :)

the problem with regexp parsing

We're using an actual PHP parser to decide whether the statement entered is "complete" or not, and in general, it works _really_ well. For example, here's what the current implementation does if you use a multi-line string with double quotes:

doublequote

And here's what it does with HEREDOC strings:

heredoc

Usually, when a statement is incomplete, PHP's parse error is "Unexpected EOF". PsySH detects when your input has that syntax error and lets you keep typing on the next line. For some reason PHP doesn't seem to use the same parse error with single quoted strings:

unexpected eof

So in this case, the correct answer is to make the current "is this statement finished?" logic detect PHP's weird response and treat it like it does the "Unexpected EOF" syntax errors.

bobthecow:
My initial shot at it seems to work:

now with more multiline awesome

… but I am worried about potential side effects. I'm trying to find other strings that have an "unexpected T_ENCAPSED_AND_WHITESPACE" syntax error, because if there are any, this will treat them like an unclosed multi-line string, and that'll be confusing to people.

bobthecow:
If we treat "unexpected T_ENCAPSED_AND_WHITESPACE" errors just like we treat "unexpected EOF" (i.e. if we don't throw a parse error, and just let you keep typing on the next line) we get this weird case:

function string

There's no way letting you keep typing input after you've entered `function 'what` will ever result in valid code, so it should be throwing an error like it does when you type `function "what` … but because of the exception we added letting you keep typing if you have an unclosed single-quoted string, it's acting as if you might possibly type something that'll make up for the fact that you started out with completely invalid code :)

bobthecow:
Annnd a workaround for a workaround :)

workaround for a workaround

We're getting somewhere, but I'm still not super comfortable about it. I'll keep playing :)

bobthecow:
Okay. My fix (along with quite a few test cases) is in the `fix/multiline-single-quote-string` branch, if you want to check it out :)

I'm still not completely happy, but it seems to be better than the current multiline string weirdness, at the very least.

Volter9:
Wow, thanks for such detailed response, I will check it out :+1:

bobthecow:
Got it! See the latest in `fix/multiline-single-quote-string`. It now supports multiline strings and also detects false positives on the parse error message :)

bobthecow:
Here's the full fix (along with a bunch of test coverage): https://github.com/bobthecow/psysh/compare/fix/multiline-single-quote-string

Volter9:
Thanks! Is it possible to install this branch globally via composer?

P.S.: The status text in your terminal looks amazing, is that custom app or custom dot files?

bobthecow:
You can:

```json
"psy/psysh": "dev-fix/multiline-single-quote-string"
```

> P.S.: The status text in your terminal looks amazing, is that custom app or custom dot files?

It's a [fish shell](http://fishshell.com) theme called [bobthefish](https://github.com/bpinto/oh-my-fish/blob/master/themes/bobthefish/README.md), which you can install via [oh-my-fish](https://github.com/bpinto/oh-my-fish/).

Volter9:
Thanks again! :smile: 
Then I can close this PR?

bobthecow:
It'll automatically close when the feature branch is released. I prefer to keep 'em open until then :)

bobthecow:
Okay. This is in develop, and will be in v0.5.0 (hopefully released shortly?)

Volter9:
Nice! Thank you again! :smile:

PR #183, opened 2015-03-21
Fix dump command

nicolas-grekas:


bobthecow:
Thanks! This has been merged into `develop` and will go out in the next release.

Issue #165, opened 2015-02-26
Presenters vs dump (and VarDumper)


barryvdh:
As I understand it, Presenters are a way to present an object in a simple way, right? Which is pretty cool because if (in Laravel for example) just do `$app = app()` the shell always displays the object. And that would be pretty large without the Presenter.

But when I run `dump $app`, shouldn't that give me more information? Eg, private properties, methods etc etc.

A cool way to do this would be the new Symfony VarDumper component: http://symfony.com/doc/current/components/var_dumper/introduction.html (and http://symfony.com/doc/current/components/var_dumper/advanced.html for some more customisation).

It also already has a CLI Dumper but could also dump to plain text. And doesn't have any external dependancies.

This is a better alternative then var_dump(), because it handles recursion/depth better, and is a bit prettier. You can try this in Laravel/Symfony with just calling `dump($var)`, but I would think that the dump command should be more like that. 

(And you can always just use the presenter by just calling the variable, without `dump`)

Markcial:
have you tried to do a `ls -la $object`?

barryvdh:
Okay did not know that. That's useful, but not quite the same.

Markcial:
why? it shows public private or protected properties or mehods as constatns.

![screenshot from 2015-02-26 12 04 45](https://cloud.githubusercontent.com/assets/208523/6390803/c0a3a0f2-bdaf-11e4-93e0-08fd6786c11e.png)

barryvdh:
Yes but it doesn't give the current state and nested properties etc, right?
The `dump` method is just a different thing then the `list` command.

barryvdh:
I think it's probably 2 things

 - The dump command uses the Presenters, so you can't use a Presenter to give a simplified version of an object and still dump all information.
 - Instead of using the default Object dumper, the VarDumper could be used for deeper/more robust information.

Markcial:
I sincerely don't see where is the trouble, the list command would fit for all of your needs, if you think that the feature needs more completion you are free to collaborate with a PR, but IMHO the list command is enough. Maybe the repo owner is able to help you better with your issue.

barryvdh:
Well I'm okay with submitting a PR myself but would like some feedback first.

bobthecow:
Check `dump -h` :)

 * `dump -a $object` dumps private and protected properties.
 * `dump --depth 100 $object` dumps deeper than the default 10. Be careful with this, as it doesn't do recursion checking :P

[...]

PR #184, opened 2015-03-21
Use VarDumper as Presenter

### LINKED ISSUE #180: Replace Presenter system by VarDumper
Hi,

I'm not sure if you know about symfony/var-dumper, so basically it's a rock solid variable presenter (using your "presenter" vocabulary).
Documentation and examples are avail. in the doc:
http://symfony.com/doc/current/components/var_dumper/index.html

I wonder if you might be interested in replacing your Presenter module by VarDumper?

nicolas-grekas:
Quite straightforward in fact! Could close #180
This gave me some ideas to enhance VarDumper that should be in the upcoming 2.7 version.
I'll also add a caster for MongoCursor and we can then work together to enhance VarDumper to the benefit of both the PsySH and Symfony Components users.
Still not interested? :)


GrahamCampbell:
I think we'd have to wait until symfony 2.7.0 was released before using this?

nicolas-grekas:
Waiting for 2.7 looks a good idea: it's coming fast, and if there is interest in merging some version of this PR, we still have a bit of time if anything needs to be ported from here to var-dumper.

nicolas-grekas:
This PR relies on symfony/symfony#14058 to be merged now: it makes integration even easier

nicolas-grekas:
@bobthecow Symfony 2.7 is beta-1 now, and the patches that ease this integration are merged. Technically we can move on. Can you please tell me if your concerns still remain? This PR taken as a PoC demonstrates that the integration is straightforward. The benefit for PsySH would be (non restrictive list): better references handling, bigger community, longer list of custom casters/presenters for all kind of objects, wider standardization around __debugInfo() and casters, more code sharing in general, etc.
I'm in SF at the end of may, I'd be happy to meet you there if you like btw!

GrahamCampbell:
If we add this, then it means psysh won't work on L5.0 anymore. People using Laravel 5.0.x will be stuck with an old version of psysh. That said, I still like this, I just don't think this should go into psysh for another 8 weeks at least.

nicolas-grekas:
We don't need to rush, we can of course take our time. I looked at Laravel source code, I see no blocker for compatibility, but we should ensure that's the case of course.

GrahamCampbell:
@nicolas-grekas No, the blocker is that laravel 5.0.x forces the 2.6.x symfony component so composer won't allow them to be installed together, so will select an earlier version of psysh.

GrahamCampbell:
Would be cool to get this into psysh 0.5.0. :)

GrahamCampbell:
Laravel 5.0 will be dropped in a few days time and 5.1 needs symfony 2.7, so we're in business.

bobthecow:
@GrahamCampbell I had a great chat with @nicolas-grekas, and there's a chance we'll replace presenters entirely with VarDumper in 0.5.0... When is Laravel 5.1 slated to land? If we do switch to VarDumper, it'd be good to get PsySH 0.5.0 in it, and swap out the Presenters for the new shiny.

GrahamCampbell:
Symfony will land today or tomorrow I think, and laravel is due about a week later, or there abouts (I'm not able to reveal the actual release date, sorry).

nicolas-grekas:
I just finished replacing presenters by VarDumper, PR is ready for feedback.
I'm now on Laravel to see what should be ported there.

bobthecow:
@nicolas-grekas Thanks for putting so much work into this. It's looking really good. A few more feedbacks… I'm not sure if any of these belong in the "psysh opinionated dumper" code, as they'd make it inconsistent with the way VarDumper works everywhere else, so maybe they're improvements that can be made to VarDumper itself?

1. In my opinion, the closer things are to standard PHP code, the easier it is to understand them at a glance, as it doesn't require (as much) context switching. As an example, it might make sense to use HEREDOC instead of triple quoted strings for multi-line string dumping. Other things are less cruicial, but also mismatched with standard PHP representations, like the `(string)` representation of a resource, which uses `#165` vs the VarDumper `@165` representation.

2. On the same note, it feels weird for array elements not to be separated by commas. It would feel more consistent for object properties to have commas after them as well, but it seems like there's a valid case either way… with array elements, though, missing commas just seems wrong.

3. VarDumper highlights _everything_, which feels like overkill. It would be easier to understand at a glance, for example, if `:`, `=>`, `[` and `{` were printed in the default terminal color, and if the quotes on strings were the same color as the string but unbolded.

4. Per PEP-8 (the canonical python coding standard, and the community convention of the only language I'm familiar with that uses triple quoted strings) the start of the string is on the same line as the opening quotes, the lines in the middle are aligned with the start of both sets of triple quotes, and the quotes at the end are on their own line:

   ```
   """this
   is
   a
   triple
   quoted
   string
   """
   ```

   VarDumper does it exactly backwards of each of those rules :)

   ```
   """
      this
      is
      a
      triple
      quoted
      string"""
   ```

nicolas-grekas:
@bobthecow thanks for the feedback.
The symfony doc has examples for most types of representations, you should review them:
http://symfony.com/doc/current/components/var_dumper/introduction.html#dump-examples-and-output

Here are some though about your points.

1. to me, the representation needs to be as straightforward as possible, but also as accurate as possible. The requirement to match the PHP syntax is already kind of included, with of course differences since we need to express things that are not possible in PHP (like objects representations). But going as far as using the heredoc syntax is not a good idea to me, because e.g. being really heredoc means putting the closing token on the first column without any indentation. Ugly IMO. But then anything else is not heredoc anymore, thus we can choose something more appropriated for humans only. For the string representation of resources, here is the reasoning: two kind of "objects" have an internal reference handle: php-objects and php-resources. Their ref-numbers live in a different "namespace". I choose `#...` for objects and `@...`. Could be swapped, but since refs are really less frequent than objects, I'm not sure this is required.

2. what's would be the purpose of the ending coma, for humans? Do you really think they help with readability? To me, they were just making dumps larger by one column. We could add them back easily.

3. the color scheme is really a configurable thing. I started var-dumper with an other one of mine, then someone else proposed to use the symfony-doc color scheme. This is the default and it won't change in Symfony of course, but can be changed trivially in psysh's dumper (although I did not semantically isolate enclosing quotes from all other surrounding characters, so your idea to style them specially needs a small code specialization).

4. this is not Python, I prefer my way for multi lines strings :-) (we could remove the indentation when the string is not nested in a hash if you really think it's wrong). In Python, triple quoted strings can be single-line strings, which kind of makes it a requirement to start it on the same line than the opening quotes. In dumps, this is really used only for multi-lines string, so we for sure know that there will be a break. Breaking just after the opening quotes makes the line shorter, which is good on a constrained console. For the closing quotes, dealing with EOL terminated strings is not infrequent. The Python style forces having what appears as a blank line for this kind of content. Thus the way it is now. More compact, still perfectly readable.

Please take all the above comments as contributions to the discussion, and by no mean as a rebuttal of yours :-) The current state of the dumped representation has already some feedback driven changes embedded (the color scheme, the `$length` displayed next to arrays, etc.) There is a big part of personal taste in any dumper's output and I really like that var-dumper here has a chance to be challenged for the better.

src/Psy/VarDumper/Presenter.php
```diff
@@ -0,0 +1,135 @@
...
+
+
+    public function __construct(OutputInterface $output)
+    {
+        $fmt = $output->getFormatter();
+        $styles = array('default' => 0);
+        foreach ($this->styles as $k => $style) {
+            $style = $fmt->getStyle($style)->apply('') ?: "\033[0m";
+            $styles[$k] = substr($style, 2, strpos($style, 'm'));
+        }
```
bobthecow: Wow, that's a pretty clever way of handling this :)

src/Psy/VarDumper/Presenter.php
```diff
@@ -0,0 +1,135 @@
...
+        "\0*\0line",
+        "\0Exception\0previous",
+    );
+    private $styles = array(
+        'num' => 'number',
+        'const' => 'const',
+        'str' => 'string',
+        'cchr' => 'string',
+        'note' => 'class',
+        'ref' => 'aside',
```
bobthecow: I think this one was better the default way… as an `<aside>` it's a bit too bright.

bobthecow:
Commas make things seem easier to parse. I'm almost never in a situation where an extra column is problematic. But I am occasionally in a situation where it's not easy to tell where array elements begin or end, especially as things get more complex. This is contrived, but I feel like commas would make this clearer:

screen shot 2015-06-13 at 5 29 49 pm

How 'bout we split the difference and separate array elements with a dark grey comma, so it's not very visible?

The triple quoted string formatting does rub me wrong. We're appropriating a feature from another language, but doing it wrong. It's like when a human language borrows a word but completely mispronounces it. Anyone who knows how it's supposed to be will be irritated. I don't want the PsySH output to be different than VarDumper, I want VarDumper to do it right ;)

Again, none of these things are blockers. I'm happy to approve this as-is. Everything from here out is just bonus awesomeness.

bobthecow:
Huh. There's something strange going on with `\t`:

screen shot 2015-06-13 at 6 01 06 pm

ogizanagi:
Actually, why were array indexes removed when dealing with sequential arrays ? 
IMO it superseded the need of comma for readability (even if I'm rather in favor of using both).

nicolas-grekas:
The code now uses the existing color scheme of psysh. I also use the OutputFormatter directly instead of this hack to extract the cli-color codes :)
Control-chars are fixed.

bobthecow:
:+1: looking good!

nicolas-grekas:
I added an experimental commit where I copy/pasted var-dumper's CliDumper with some tweaks:
- ending coma added at end of values,
- `#handle` is back near to objects and resources dumps, using the default color of the cli.
- removed ending `{}` on resources lines ; added the `resource` suffix instead of `:` prefix (so that `stream resource #123` is displayed now instead of `:stream {}`)
- I changed the logic for multi-lines string:
  - if the string is not longer than 80 characters (CRLF included), then it is dumped on a single line,
  - for longer multi-lines string, I now always put the ending `"""` on its own line (same for the opening one), and I made the `\n` char visible at each end of line. Do you like it?

nicolas-grekas:
Once we agree on a final dump, I'll submit a PR on symfony with the updated CliDumper.

nicolas-grekas:
capture du 2015-06-16 11 36 54

bobthecow:
* Yay commas! :)

* :+1: new way resources and refs are presented, because of the reasons I stated before (not having to context switch, being more familiar, etc). It feels like an improved version of `(string) $resource` and `var_dump($obj)`, in ways which aren't surprising, which has always been my goal with dumping things… That said, if you strongly prefer `@123` for refs and the `:stream` style annotation,  I don't think it's a problem to keep it the way it was. Using two different glyphs for the ids makes it a bit more clear that a resource is most definitely not an object.

* I hadn't thought of it, but I do think I like showing newlines explicitly in `"""` strings. One of the things PsySH stole from Fish shell is that when stdout doesn't end with a newline (so it has to add one implicitly) it lets you know, so it's very obvious the difference between `echo "foo\n"` and `echo "foo"`. Especially with non-visible characters, :+1: for making them visible.

* Style-wise, I liked the inverted black-on-green color highlighting for control characters that we had before.

* If a shell has 256 color support, it seems like it might be a win to use `38;5;247` (the color VarDumper uses for refs) for things like refs and commas rather than default, to make those less-important things a bit less visually strong (and maybe `=>`? maybe even for the `+` and such on property names?).

* I don't have a strong preference on short strings with newlines being shown on one line or not. I think I'd lean _slightly_ toward one line like you changed it to, because it's a bit more concise, and solves for the not-uncommon case of having a single-line string that ends with a newline (`"foo\n"`).

nicolas-grekas:
For the color differences you suggest (`#123`, `=>`, etc.), I prefer using the default cli foreground color. Our brain reads the default foreground color as the "no-special-color" one. At least that's how I feel about it: everything in the default color is less important, not colored, no special semantic. The second argument is about people with a white background CLI. Using the default foreground will just swap white for black and make things as readable in both modes.

For the video inversion you liked on control chars, I removed it when I added the explicit end of lines. A colored background is very flashy and I didn't what to impose that to everyone doing multi lines strings (which is a lot of people, more that with the previous pretty rare control chars).

Are you ok to keep the color as is?

I don't know what will be accepted in Sf 2.7 as bug fix and what will have to be in 2.8.
That means that we have two choices to continue:
- keep the tweaked CliDumper in psysh's code for now until 2.8 is out
- accept the differences for now, knowing that once 2.8 is out, everything will be enhanced (ending `"""` on the same line, no ending coma, `:stream` style and some other missing tweaks you suggested)

What is your choice? Both are fine by me.

nicolas-grekas:
See symfony/symfony#15036
After having asked for a few more feedbacks, I reverted the single-lining of short strings but kept it for single-line strings that ends with a newline.

bobthecow:
:+1: That seems like an entirely reasonable middle ground.

nicolas-grekas:
The PR on Symfony is now merged, and this one is ready to be merged also IMHO :)

GrahamCampbell:
Should we bump the symfony version first?

GrahamCampbell:
`"symfony/var-dumper": "^2.7.2|~3.0",`

nicolas-grekas:
It is not technically required: it works on 2.7.0/1, it's just not as pretty as it will be once 2.7.2 is tagged

GrahamCampbell:
Then as soon as symfony v2.7.2 is tagged, then we're in a position to do a develop->master merge here.

GrahamCampbell:
I still think we should enforce the pretty version. :)

nicolas-grekas:
But then you will prevent people from using psysh with a 2.7.0/1 symfony :(

GrahamCampbell:
Yeh, true. But who would want to use that after 2.7.2 is out?

GrahamCampbell:
Meh, tbh, I'm not massively fussed. Mergy mergy!

bobthecow:
Yeah, I tend to err on the side of supporting the minimum version possible for whatever our combination of dependencies allows, as long as that's not too arduous for us to maintain. In this case, there's no difference at all in our project… might as well support the ugly version just in case someone's stuck on 2.7.0 for a bit :)

bobthecow:
It does, however, mean that PsySH requires PHP 5.3.9 now. Do you mind updating composer.json for that?

GrahamCampbell:
I think it would make sense for us to bump to 5.4+ plus. 5.3 isn't maintained.

GrahamCampbell:
If not, deffo 5.5.9+.

GrahamCampbell:
Typo there, meant 5.3.9 not 5.5.9...

bobthecow:
It's no harder for us to support 5.3.9 than 5.4+. It makes sense for all our *users* to update, but see two comments ago re: supporting older versions :)

GrahamCampbell:
k :)

bobthecow:
It'd be awesome if Packagist could give package maintainers insight into PHP and dependency versions that package users are running. I'd feel better about dropping support for older versions if I knew how many people I'd be abandoning.

GrahamCampbell:
> It'd be awesome if Packagist could give package maintainers insight into PHP and dependency versions 

They did a blog post a few months back actually.

GrahamCampbell:
http://seld.be/notes/my-view-of-php-version-adoption

bobthecow:
I mean on a per-package basis. For example, I'd guess that Mustache.php users are running older versions than PsySH users, on average, if for no other reason than several WordPress plugins and themes use Mustache.php.

GrahamCampbell:
I don't think it really matters what package maintainers do. Surely what's of more interest is the version of php real people are actually using?

nicolas-grekas:
PR updated 5.3.9

bobthecow:
Right. That's what I'm saying. For this project, I care how many PsySH users would be unable to update if I dropped support for 5.3.x. For another project, I'd care about how many users of _that_ project would be affected. In my example, I'm almost positive that dropping 5.3.x support in Mustache.php would affect a significant number of mustache users.

GrahamCampbell:
:+1:

GrahamCampbell:
@bobthecow What are we going to do about StyleCI? Should we disable the preincrrementation fixer?

GrahamCampbell:
@bobthecow Hmmm. Yeh, that would be useful. I suspect 90% of the installs of psysh are due to laravel though?

bobthecow:
Also, because of Ubuntu LTS I'm super reluctant to drop support for 5.3. It sucks, but there are a lot of Ubuntu servers out there. As it is, requiring 5.3.9 abandons anyone running CentOS LTS who isn't compiling their own PHP.

bobthecow:
It's a significant portion, but way less than 90% ;)

Laravel has 3233 installs today, PsySH has 7400. Monthly installs have about the same breakdown. And Laravel has been updated quite a bit more often than PsySH, so the version updates are probably skewing the numbers even more toward Laravel.

bobthecow:
Okay, I'll merge this as is. I'm open to suggestions on the StyleCI thing, but I'm not going to make @nicolas-grekas fix it, because he didn't break it.

bobthecow:
:boom:

bobthecow:
Oh, I might be wrong about that. I just checked `laravel/laravel`, not `laravel/framework`. The framework has 12k installs, so the version updates are _really_ skewing the numbers. It's hard to say :)

nicolas-grekas:
Yeehaaaa! That needs a tweet :-)

GrahamCampbell:
laravel/laravel doesn't get many installs. laravel/framework is the one that'll have MANY installs per month. :)

GrahamCampbell:
How does https://github.com/bobthecow/psysh/pull/209 look?

bobthecow:
@nicolas-grekas Wanna take a shot at a PR for ezsh? I think that's the only other project with custom presenters.

https://github.com/lolautruche/ezsh/tree/master/src/Presenter

nicolas-grekas:
I will thanks for the pointer

nicolas-grekas:
PR for ezsh is here: https://github.com/lolautruche/ezsh/pull/3

bobthecow:
Thanks!

lolautruche:
@bobthecow Many thanks for this hint :smiley:

bobthecow:
Of course!

PR #186, opened 2015-03-26
Initial support for php 7

### LINKED ISSUE #185: PHP 7-dev Error
This may a bit early, but better sooner than later. When trying to start psysh in PHP7-dev, you see:

```
Fatal error: "Scalar\String" cannot be used as a type declaration in phar:///app/php/bin/psysh/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.php on line 86
```

GrahamCampbell:
Closes #185.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,8 @@
     ],
     "require": {
         "php": ">=5.3.0",
-        "symfony/console": "~2.3.10|~2.4.2|~2.5",
-        "nikic/php-parser": "~1.0",
+        "symfony/console": "~2.3.10|^2.4.2",
+        "nikic/php-parser": "^1.2.1",
         "dnoegel/php-xdg-base-dir": "0.1",
         "jakub-onderka/php-console-highlighter": "0.3.*"
     },
@@ -40,7 +40,7 @@
     "bin": ["bin/psysh"],
     "extra": {
         "branch-alias": {
-            "dev-develop": "0.4.x-dev"
+            "dev-develop": "0.5.x-dev"
         }
     }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
   - hhvm-nightly
 
@@ -16,6 +17,7 @@ script:
 
 matrix:
   allow_failures:
+    - php: 7.0
     - php: hhvm
     - php: hhvm-nightly
   fast_finish: true
--- a/src/Psy/CodeCleaner/MagicConstantsPass.php
+++ b/src/Psy/CodeCleaner/MagicConstantsPass.php
@@ -16,7 +16,7 @@ use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Name;
 use PhpParser\Node\Scalar\MagicConst\Dir;
 use PhpParser\Node\Scalar\MagicConst\File;
-use PhpParser\Node\Scalar\String as String;
+use PhpParser\Node\Scalar\String_ as StringNode;
 
 /**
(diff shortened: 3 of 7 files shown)


GrahamCampbell:
Has anyone got any idea what's going on with the tests here?

bobthecow:
Interesting. Apparently PHP Parser (as of 1.2.0) now rewrites things based on the `use` statements? The changelog mentions a "use builder" and "namespace builder", and I imagine they're the culprits.

The tests should prolly support both ways, since PsySH works (and will continue to work) with older versions of PHP Parser.

bobthecow:
Or maybe it's that PHP-Parser's use statement parsing is no longer compatible, so the UseStatement CodeCleaner pass isn't able to rewrite use statements. I bet that's it.

GrahamCampbell:
@bobthecow Though, that means our current release is broken though - we're using 1.x, so we're getting all changes from 1.2.x.

bobthecow:
In other words, 1.2 is a BC break? Sweet.

GrahamCampbell:
@bobthecow It's not meant to be? Ping @nikic.

bobthecow:
The problem with this is that we have to yank all previous releases of PsySH to keep things from resolving to an older version of PsySH if there's a dependency conflict. I wish people were better at SemVer :-/

GrahamCampbell:
@bobthecow Give @nikic some time. It's possible we can fix this in the next patch release of php parser.

nikic:
The problem is likely the check for `Traversable` in https://github.com/bobthecow/psysh/blob/master/src/Psy/CodeCleaner/UseStatementPass.php#L78 - nodes are no longer `Traversable` (but they can still be traversed, like any other object).

You should be able to safely drop that line - it doesn't make sense anyway, as you already typehint against `Node`.

bobthecow:
@nikic Awesome. Tests pass when that check is replaced with an `else`.

GrahamCampbell:
I'll send a pull fix that separately then.

bobthecow:
:+1: Awesome. Thanks all. @nikic, thanks for the quick reply.

GrahamCampbell:
@bobthecow https://github.com/bobthecow/psysh/pull/188

GrahamCampbell:
Also, that php7 thing needs a new minor release rather than a patch release since we're renaming classes.

GrahamCampbell:
Alrightly. How does this look now?

composer.json
```diff
@@ -14,8 +14,8 @@
     ],
     "require": {
         "php": ">=5.3.0",
-        "symfony/console": "~2.3.10|~2.4.2|~2.5",
-        "nikic/php-parser": "~1.0",
+        "symfony/console": "~2.3.10|^2.4.2",
+        "nikic/php-parser": "^1.2.1",
```
bobthecow: How long ago did the carat operator ship? I checked the composer changelog but didn't see it.

GrahamCampbell: Ages ago now. Probably over 2 months. Let me check for you...

bobthecow:
You know, if we're going to do a minor version bump, I'd like to see if we can get some of the other PRs into it. Maybe we ship a point release with the PHP-Parser fix first, then get this out after?

GrahamCampbell:
@bobthecow Fine with me. :)

bobthecow:
:+1: this PR looks good though.

composer.json
GrahamCampbell: It was 8th December 2014: https://github.com/composer/composer/pull/2596.

bobthecow: :+1: Sweet. Works for me.

bobthecow:
v0.4.4 is out with the initial fix, I'll merge this one into develop and we'll get started on v0.5.0 changes :)

PR #187, opened 2015-03-26
Fixed incorrect phpdoc

GrahamCampbell:


bobthecow:
Thanks! Merging this despite the (unrelated) build failure :)

Which reminds me, one of these days we should update to use typed array notation instead of just `array`.

GrahamCampbell:
Yeh, probably. :)

PR #205, opened 2015-06-17
Don't try and match where we've not been able to find the variable

davedevelopment:
At present, this includes

``` php
$app['something']->
$app->getSomething()->
```
etc.

It would be great to be able to parse back far enough, but I think that's quite a task

### DIFF
--- a/src/Psy/TabCompletion/Matcher/ObjectAttributesMatcher.php
+++ b/src/Psy/TabCompletion/Matcher/ObjectAttributesMatcher.php
@@ -11,6 +11,8 @@
 
 namespace Psy\TabCompletion\Matcher;
 
+use InvalidArgumentException;
+
 /**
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Wrong branch. Needs to go to develop.

davedevelopment:
@GrahamCampbell see #206. Is the project following gitflow or just a new stuff goes in develop type thing. I did check the `CONTRIBUTING.md` before the PR, nothing in there, happy to PR something for that, just need to know what the rules are.

bobthecow:
This project doesn't really follow git flow, since it's probably less applicable to open source projects than to teams. We don't really care where you do your work, as long as pull requests for all pre-release code end up in `develop`.

And yeah, it'd prolly make sense to add that to CONTRIBUTING. I thought it was already there, TBH.

PR #206, opened 2015-06-17
Don't try and match where we've not been able to find the variable

davedevelopment:
At present, this includes

```
$app['something']->
$app->getSomething()->
```

etc.

It would be great to be able to parse back far enough, but I think that's quite a task

### DIFF
--- a/src/Psy/TabCompletion/Matcher/ObjectAttributesMatcher.php
+++ b/src/Psy/TabCompletion/Matcher/ObjectAttributesMatcher.php
@@ -41,7 +41,7 @@ class ObjectAttributesMatcher extends AbstractContextAwareMatcher
         try {
             $object = $this->getVariable($objectName);
         } catch (InvalidArgumentException $e) {
-            return [];
+            return array();
         }
 
         return array_filter(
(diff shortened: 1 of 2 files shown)


bobthecow:
Talk to me a bit more about this.

davedevelopment:
Sure thing! That `hasMatched` method essentially checks to make sure either the last or second to last token was a `T_OBJECT_OPERATOR`, which is fine. However, once we're up in to the `getMatches` method, the code assumes that whatever comes before the `T_OBJECT_OPERATOR` is both a `T_VARIABLE` and it will be available in the current variable context. 

So when I type

``` php
$app['some.pimple.service']-><tab><tab>
```

This matcher would attempt to call `getVariable` with an empty string, because the token before the `T_OBJECT` is a `]`. `getVariable` throws if it can't find whatever variable it is that you ask for.

We could add code to `hasMatched` as well to prevent things getting that far, at least for the cases where it's not a `T_OBJECT`, but I think it's safest to keep this catch in here. I'd rather not have any matches shown than see an exception.

bobthecow:
Yeah, that makes sense.

Do you mind switching to PHP 5.3-style arrays so this passes CI? :)

davedevelopment:
No problem, didn't even notice the failures :/

GrahamCampbell:
Rebasing should allow styleci to pass.

bobthecow:
@GrahamCampbell I'll merge without a rebase. I know they're not his errors :)

PR #208, opened 2015-06-19
Added PsyshBundle

theofidry:


bobthecow:
Thanks for adding this!

I've kept the list of projects for frameworks with multiple integrations in chronological order, with the oldest implementations first. Move this one to the end of the list and I'll merge :)

theofidry:
Actually I would like to have your opinion on this: there is currently 3 bundles for PsySH integration in Symfony:
* navitronic/psymf
* varas/sf2-psysh-bootstrap
* luxifer/symfony-repl

Before making my own, I checked the 3 others. Out of them, only `navitronic/psymf` seemed to provider a real integration with Symfony as a Bundle. The others are simple php scripts working with require on bootstrap and kernel and such. It works, but it's not a bundle.

So out of the 3 working, only `navitronic/psymf` is designed as a Symfony bundle, which IMO is the best integration for this framework. First I planned to do a fork and merge back my changes to that project. As its author was not very reactive and that I did lots of work on my own to, I'm keeping my fork as it is.

My point is, should we really mention the 4 of them? Since 2 of them are not real integration on Symfony and the 2 others are quite the same, 1 active which refer to the non active, wouldn't just mentioning one be enough?

bobthecow:
They're still all integrations, even if they're not bundles ;)

In time, one of these integrations will likely become the de facto way to do it, and we can recommend that. Until then, I'd rather keep links to older or less awesome projects than remove 'em preemptively.

theofidry:
done :)

bobthecow:
Thanks!

theofidry:
@bobthecow when will this be merged to the master branch? Since it's really just a part of the doc and not related to any new feature, wouldn't it be possible to cherry pick this commit?

bobthecow:
Nope. master is always a stable release in this repo. `v0.5.0` will be out very soon, though (likely this weekend?)

theofidry:
Ok great!

Issue #222, opened 2015-08-04
StyleCI


GrahamCampbell:
You seem to have revoked our access token @bobthecow?

bobthecow:
Everything should be wired up and running again. Any idea why this one fails on StyleCI but work (with equivalent config) locally?

https://styleci.io/analyses/86ngn8

GrahamCampbell:
Probably because we don't use the same version of php-cs-fixer that's public. Ours is better. :P

GrahamCampbell:
This looks to be psr2 related though.

GrahamCampbell:
PSR2 doesn't allow those spaces there.

bobthecow:
PSR2 is a jerk.

GrahamCampbell:
You could turn off the function argument fixer?

bobthecow:
I re-read that section of PSR-2, and I think I disagree.

> In the argument list, there MUST NOT be a space before each comma, and there MUST be one space after each comma.

I read this as "…there MUST be (at least) one space after each comma." Whoever wrote that check reads it as "…there MUST be (exactly) one space after each comma." You could make an argument about what was intended by that statement, but it does allow both interpretations. And when a standard is ambiguous, it has to allow all reasonable interpretations.

Where do I open a bug report for this argument fixer? :)

GrahamCampbell:
> In the argument list, there MUST NOT be a space before each comma, and there MUST be one space after each comma.

That means exactly one. It's just badly worded. If it meant one more more, I think it would have said "more than one".

bobthecow:
I disagree. Anyway, it's kind of a moot point, since it's highly unlikely they'll change it, and it's highly unlikely either of us will convince the other that their interpretation is correct. Closing for now ;)

PR #227, opened 2015-08-21
Fixed CS Config

GrahamCampbell:


GrahamCampbell:
There you go @bobthecow. Sorry I didn't get to this earlier. :)

bobthecow:
Oh, nice. I didn't know about `long_array_syntax`.

I'll merge this for now, but I think the `method_argument_space` fixer is wrong. See [my comment on #222](https://github.com/bobthecow/psysh/issues/222#issuecomment-133878447) :)

GrahamCampbell:
I think it's right. Maybe you need to make an issue for the PSR team to look at?

GrahamCampbell:
Only they can collectively confirm what that rule actually meant, then we can take it from there. :)

bobthecow:
It doesn't actually matter what they meant. Unless they update the spec (which they almost certainly won't) then it's ambiguous and both readings are valid.

GrahamCampbell:
> Unless they update the spec (which they almost certainly won't) then it's ambiguous and both readings are valid.

I agree it's ambiguous, but I don't agree that the spec can't change. "Bug fixes" are surely a good thing, and the specs aren't never changed. As I said, it's not for me to say. You'd have to ask on the mailing list or something. :)

bobthecow:
I didn't mean the spec can't change. I meant the spec almost certainly won't. I've been around the FIG long enough to realize that :P

GrahamCampbell:
> I meant the spec almost certainly won't.

Lol, yeh, I know what you mean. It still might be worth getting some clarification though.

PR #228, opened 2015-08-21
Add support for local config files

shadowhand:
Allows creating `.psysh.php` file that will overload global configuration on a per-project basis.

cc @ameech

### DIFF
--- a/src/Psy/Configuration.php
+++ b/src/Psy/Configuration.php
@@ -105,6 +105,9 @@ class Configuration
      * This checks for the presence of Readline and Pcntl extensions.
      *
      * If a config file is available, it will be loaded and merged with the current config.
+     *
+     * If no custom config file was specified and a local project config file
+     * is available, it will be loaded and merged with the current config.
      */
     public function init()
     {
@@ -114,6 +117,10 @@ class Configuration
 
         if ($configFile = $this->getConfigFile()) {
             $this->loadConfigFile($configFile);
+        }
+
+        if (!$this->configFile && $localConfig = $this->getLocalConfigFile()) {
+            $this->loadConfigFile($localConfig);
         }
     }
 
@@ -140,13 +147,30 @@ class Configuration
         foreach ($this->getConfigDirs() as $dir) {
             $file = $dir . '/config.php';
             if (@is_file($file)) {
-                return $this->configFile = $file;
+                return $file;
             }
 
             $file = $dir . '/rc.php';
             if (@is_file($file)) {
-                return $this->configFile = $file;
+                return $file;
             }
+        }
+    }
+
+    /**
+     * Get the local PsySH config file.
+     *
+     * Searches for a project specific config file `.psysh.php` in the current
+     * working directory.
+     *
+     * @return string
+     */
+    public function getLocalConfigFile()
+    {
+        $localConfig = getenv('PWD') . '/.psysh.php';
+
+        if (@is_file($localConfig)) {
+            return $localConfig;
         }
     }
 
--- a/src/Psy/functions.php
+++ b/src/Psy/functions.php
@@ -34,6 +34,7 @@ if (!function_exists('Psy\info')) {
             'error logging level' => $config->errorLoggingLevel(),
             'config file'         => array(
                 'default config file' => $config->getConfigFile(),
+                'local config file'   => $config->getLocalConfigFile(),
                 'PSYSH_CONFIG env'    => getenv('PSYSH_CONFIG'),
             ),
             // 'config dir'  => $config->getConfigDir(),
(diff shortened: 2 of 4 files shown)


shadowhand:
@GrahamCampbell correct branch I think, and I fixed the 5.3 regresssion.

bobthecow:
The local config file should only be loaded if an explicit config file was not provided. For example, 

```bash
psysh --config myconfig.php
# or
PSYSH_CONFIG=myconfig.php psysh
```

… skips the global config file in favor of an alternate config file. If PsySH was invoked like this, we don't want to merge in `.psysh.php` on top of the explicit config.

src/Psy/Configuration.php
```diff
@@ -140,17 +147,34 @@ public function getConfigFile()
...
             $file = $dir . '/config.php';
             if (@is_file($file)) {
-                return $this->configFile = $file;
+                return $file;
             }
 
             $file = $dir . '/rc.php';
             if (@is_file($file)) {
-                return $this->configFile = $file;
+                return $file;
```
shadowhand: @bobthecow this seemed like the cleanest way to ensure that checking `$this->configFile` would only be set by the user providing a config file. It does have a (small) performance implication if the user calls `getConfigFile` more than once. By default there is only one invocation, so it seemed safe to change.

src/Psy/Configuration.php
```diff
@@ -115,6 +118,10 @@ public function init()
         if ($configFile = $this->getConfigFile()) {
             $this->loadConfigFile($configFile);
         }
+
+        if (!$this->configFile && $localConfig = $this->getLocalConfigFile()) {
```
shadowhand: @bobthecow now skips loading local config if the user passed a config file option.

src/Psy/Configuration.php
bobthecow: :+1: makes sense.

src/Psy/Configuration.php
bobthecow: I'd like to see a test case to assert this, just to keep us from making a change that regresses in the future, once we don't have the context of this pull request :)

bobthecow:
Do you mind adding this to Psy\info as well?

https://github.com/bobthecow/psysh/blob/master/src/Psy/functions.php#L35

bobthecow:
The StyleCI issue should go away if you rebase onto the current develop branch.

src/Psy/Configuration.php
shadowhand: Test case has been added ...

test/Psy/Test/ConfigurationTest.php
```diff
@@ -132,6 +141,26 @@ public function testLoadConfigFile()
         $this->assertEquals(E_ALL & ~E_NOTICE, $config->errorLoggingLevel());
     }
 
+    public function testLoadLocalConfigFile()
```
shadowhand: ... here.

src/Psy/functions.php
```diff
@@ -34,6 +34,7 @@ function info()
             'error logging level' => $config->errorLoggingLevel(),
             'config file'         => array(
                 'default config file' => $config->getConfigFile(),
+                'project config file' => $config->getLocalConfigFile(),
```
shadowhand: @bobthecow info updated.

shadowhand:
> StyleCI issue should go away

Done.

test/Psy/Test/ConfigurationTest.php
```diff
@@ -132,6 +141,26 @@ public function testLoadConfigFile()
...
+    {
+        $oldPwd = getenv('PWD');
+        putenv('PWD=' . realpath(__DIR__ . '/../../fixtures/default/'));
```
shadowhand: @bobthecow maybe I should do all of the local file testing here instead of having those modifications in `testFilesAndDirectories`...?

src/Psy/functions.php
bobthecow: let's call it 'local config file', just to match how we're using it.

test/Psy/Test/ConfigurationTest.php
bobthecow: Yeah, let's do that.

test/Psy/Test/ConfigurationTest.php
bobthecow: Thanks!

test/fixtures/project/.psysh.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
shadowhand: Created a separate directory for this file so it clearly for a different test than directories.

src/Psy/functions.php
```diff
@@ -34,6 +34,7 @@ function info()
             'error logging level' => $config->errorLoggingLevel(),
             'config file'         => array(
                 'default config file' => $config->getConfigFile(),
+                'local config file'   => $config->getLocalConfigFile(),
```
shadowhand: ~~project~~ local

bobthecow:
Awesome, thanks!

PR #214, opened 2015-07-10
Fix incompatibility with Symfony\Component\Console\Application::renderException

mattsah:
Symfony added typehinting on this method at some point, and it's now raising an error.

```
Whoops\Exception\ErrorException: Declaration of Psy\Shell::renderException() should be compatible with Symfony\Component\Console\Application::renderException(Exception $e, Symfony\Component\Console\Output\OutputInterface $output) in file /home/matt/Dropbox/Projects/iw.quickstart/vendor/psy/psysh/src/Psy/Shell.php on line 43
Stack trace:
  1. Whoops\Exception\ErrorException->() /home/matt/Dropbox/Projects/iw.quickstart/vendor/psy/psysh/src/Psy/Shell.php:43
  2. Whoops\Run->handleError() /home/matt/Dropbox/Projects/iw.quickstart/vendor/psy/psysh/src/Psy/Shell.php:43
  3. include() /home/matt/Dropbox/Projects/iw.quickstart/vendor/composer/ClassLoader.php:412
  4. Composer\Autoload\includeFile() /home/matt/Dropbox/Projects/iw.quickstart/vendor/composer/ClassLoader.php:301
  5. Composer\Autoload\ClassLoader->loadClass() :0
  6. spl_autoload_call() /home/matt/Dropbox/Projects/iw.quickstart/vendor/dotink/inkwell-console/src/Quill.php:8
  7. include() /home/matt/Dropbox/Projects/iw.quickstart/vendor/composer/ClassLoader.php:412
  8. Composer\Autoload\includeFile() /home/matt/Dropbox/Projects/iw.quickstart/vendor/composer/ClassLoader.php:301
  9. Composer\Autoload\ClassLoader->loadClass() :0
 10. spl_autoload_call() /home/matt/Dropbox/Projects/iw.quickstart/bin/quill:34
 11. {closure}() :0
 12. call_user_func_array() /home/matt/Dropbox/Projects/iw.quickstart/vendor/dotink/affinity/src/Engine.php:121
 13. Affinity\Engine->exec() /home/matt/Dropbox/Projects/iw.quickstart/vendor/dotink/inkwell-core/src/Core.php:188
 14. Inkwell\Core->run() /home/matt/Dropbox/Projects/iw.quickstart/bin/quill:86
```

This fixes it.

PR #235, opened 2015-08-30
2.x and 3.0 compat with symfony/console

nicolas-grekas:
Replaces #214

### DIFF
--- a/src/Psy/Shell.php
+++ b/src/Psy/Shell.php
@@ -279,7 +279,11 @@ class Shell extends Application
             $output = $this->config->getOutput();
         }
 
-        return parent::run($input, $output);
+        try {
+            return parent::run($input, $output);
+        } catch (\Exception $e) {
+            $this->writeException($e);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
See also https://github.com/symfony/symfony/pull/15657
Even if this PR on symfony is merged, I think this one is right.

src/Psy/Shell.php
```diff
@@ -279,7 +279,13 @@ public function run(InputInterface $input = null, OutputInterface $output = null
...
         }
 
-        return parent::run($input, $output);
+        try {
+            return parent::run($input, $output);
+        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
+        }
+
+        $this->writeException($e);
```
GrahamCampbell: this could result in a fatal error

GrahamCampbell: you need to convert the throwable object to an exception first

nicolas-grekas: right, I've removed the cath for throwables, it's an other topic for an other PR

bobthecow:
Thanks @nicolas-grekas!

PR #239, opened 2015-09-09
support fetch ::class for interface or class not exists

### LINKED ISSUE #238: Could not fetch ::class consts properly
When execute `A::class`,  no matter the class `A` is exists or the `A` is an `interface`, we should get `"A"`, but given
```
PHP Fatal error:  Class 'A' not found in eval()'d code on line 1
```

In fact, there is no need to check the existance of class while fetching the `::class` const.

lvht:
try to fix #238

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,11 @@ matrix:
     - php: 7.0
     - php: hhvm
   fast_finish: true
+
+notifications:
+  webhooks:
+    urls:
+      - https://webhooks.gitter.im/e/0c7cfb83f9925542003e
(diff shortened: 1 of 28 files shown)


bobthecow:
Thanks! A few things:

 * Please develop against the `develop` branch, as that has the most recent code and is where we stage things for the next release.
 * PsySH supports PHP < 5.5 as well, so there needs to be a version check here.
 * `ClassName::CLASS` should be supported as well.
 * I'd like to see some test coverage for this. It should be as simple as adding a line or two to an array :)

PR #251, opened 2015-10-31
PHP 7 support

alexeyshockov:
This is initial support for PHP 7 (that's available in 2.0 version of great PHP parser by nikic).

I haven't found any plans for the topic, so I've decided to create a PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "php": ">=5.3.9",
         "symfony/console": "~2.3.10|^2.4.2|~3.0",
         "symfony/var-dumper": "~2.7|~3.0",
-        "nikic/php-parser": "^1.2.1",
+        "nikic/php-parser": "^1.2.1|~2.0",
         "dnoegel/php-xdg-base-dir": "0.1",
         "jakub-onderka/php-console-highlighter": "0.3.*"
     },
--- a/src/Psy/CodeCleaner.php
+++ b/src/Psy/CodeCleaner.php
@@ -11,7 +11,6 @@
 
 namespace Psy;
 
-use PhpParser\Lexer;
 use PhpParser\NodeTraverser;
 use PhpParser\Parser;
 use PhpParser\PrettyPrinter\Standard as Printer;
@@ -53,7 +52,12 @@ class CodeCleaner
      */
     public function __construct(Parser $parser = null, Printer $printer = null, NodeTraverser $traverser = null)
     {
-        $this->parser    = $parser    ?: new Parser(new Lexer());
+        if ($parser === null) {
+            $parserFactory = new ParserFactory();
+            $parser        = $parserFactory->createParser();
+        }
+
+        $this->parser    = $parser;
         $this->printer   = $printer   ?: new Printer();
         $this->traverser = $traverser ?: new NodeTraverser();
 
--- a/src/Psy/Command/ParseCommand.php
+++ b/src/Psy/Command/ParseCommand.php
@@ -11,9 +11,9 @@
 
 namespace Psy\Command;
 
-use PhpParser\Lexer;
 use PhpParser\Node;
 use PhpParser\Parser;
+use Psy\ParserFactory;
 use Psy\VarDumper\Presenter;
 use Psy\VarDumper\PresenterAware;
 use Symfony\Component\Console\Input\InputArgument;
@@ -28,7 +28,19 @@ use Symfony\Component\VarDumper\Caster\Caster;
 class ParseCommand extends Command implements PresenterAware
 {
     private $presenter;
-    private $parser;
+    private $parserFactory;
+    private $parsers;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($name = null)
+    {
+        $this->parserFactory = new ParserFactory();
+        $this->parsers = array();
+
+        parent::__construct($name);
+    }
 
     /**
      * PresenterAware interface.
@@ -59,12 +71,26 @@ class ParseCommand extends Command implements PresenterAware
      */
     protected function configure()
     {
+        $definition = array(
+            new InputArgument('code', InputArgument::REQUIRED, 'PHP code to parse.'),
+            new InputOption('depth', '', InputOption::VALUE_REQUIRED, 'Depth to parse', 10),
+        );
+
+        if ($this->parserFactory->hasKindsSupport()) {
+            $definition[] = new InputOption(
+                'kind',
+                '',
+                InputOption::VALUE_REQUIRED,
+                'One of PhpParser\\ParserFactory constants: '
+                    . implode(', ', ParserFactory::getPossibleKinds())
+                    . " (default is based on current interpreter's version)",
+                $this->parserFactory->getDefaultKind()
+            );
+        }
+
         $this
             ->setName('parse')
-            ->setDefinition(array(
-                new InputArgument('code', InputArgument::REQUIRED, 'PHP code to parse.'),
-                new InputOption('depth', '', InputOption::VALUE_REQUIRED, 'Depth to parse', 10),
-            ))
+            ->setDefinition($definition)
             ->setDescription('Parse PHP code and show the abstract syntax tree.')
             ->setHelp(
                 <<<HELP
@@ -90,22 +116,22 @@ HELP
             $code = '<?php ' . $code;
         }
 
-        $depth = $input->getOption('depth');
-        $nodes = $this->parse($code);
+        $parserKind = $input->getOption('kind');
+        $depth      = $input->getOption('depth');
+        $nodes      = $this->parse($this->getParser($parserKind), $code);
         $output->page($this->presenter->present($nodes, $depth));
     }
 
     /**
      * Lex and parse a string of code into statements.
      *
+     * @param Parser $parser
      * @param string $code
      *
      * @return array Statements
      */
-    private function parse($code)
+    private function parse(Parser $parser, $code)
     {
-        $parser = $this->getParser();
-
         try {
             return $parser->parse($code);
         } catch (\PhpParser\Error $e) {
@@ -121,14 +147,16 @@ HELP
     /**
      * Get (or create) the Parser instance.
      *
+     * @param string|null $kind One of Psy\ParserFactory constants (only for PHP parser 2.0 and above).
+     *
      * @return Parser
      */
-    private function getParser()
+    private function getParser($kind = null)
     {
-        if (!isset($this->parser)) {
-            $this->parser = new Parser(new Lexer());
+        if (!array_key_exists($kind, $this->parsers)) {
+            $this->parsers[$kind] = $this->parserFactory->createParser($kind);
         }
 
-        return $this->parser;
+        return $this->parsers[$kind];
     }
 }
--- a/src/Psy/ParserFactory.php
+++ b/src/Psy/ParserFactory.php
@@ -1,0 +1,91 @@
+<?php
+
+/*
+ * This file is part of Psy Shell
+ *
+ * (c) 2012-2014 Justin Hileman
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy;
+
+use PhpParser\Lexer;
+use PhpParser\Parser;
+use PhpParser\ParserFactory as OriginalParserFactory;
+
+/**
+ * Parser factory to abstract over PHP parser library versions.
+ */
+class ParserFactory
+{
+    const ONLY_PHP5   = 'ONLY_PHP5';
+    const ONLY_PHP7   = 'ONLY_PHP7';
+    const PREFER_PHP5 = 'PREFER_PHP5';
+    const PREFER_PHP7 = 'PREFER_PHP7';
+
+    /**
+     * Possible kinds of parsers for the factory, from PHP parser library.
+     *
+     * @return array
+     */
+    public static function getPossibleKinds()
+    {
+        return array('ONLY_PHP5', 'ONLY_PHP7', 'PREFER_PHP5', 'PREFER_PHP7');
+    }
+
+    /**
+     * Is this parser factory supports kinds?
+     *
+     * PHP parser < 2.0 doesn't support kinds, >= 2.0 — does.
+     *
+     * @return bool
+     */
+    public function hasKindsSupport()
+    {
+        return class_exists('PhpParser\ParserFactory');
+    }
+
+    /**
+     * Default kind (if supported, based on current interpreter's version).
+     *
+     * @return string|null
+     */
+    public function getDefaultKind()
+    {
+        if ($this->hasKindsSupport()) {
+            return version_compare(PHP_VERSION, '7.0', '>=') ? static::ONLY_PHP7 : static::ONLY_PHP5;
+        }
+    }
+
+    /**
+     * New parser instance with given kind.
+     *
+     * @param string|null $kind One of class constants (only for PHP parser 2.0 and above).
+     *
+     * @return Parser
+     */
+    public function createParser($kind = null)
+    {
+        if ($this->hasKindsSupport()) {
+            $originalFactory = new OriginalParserFactory();
+
+            $kind = $kind ?: $this->getDefaultKind();
+
+            if (!in_array($kind, static::getPossibleKinds())) {
+                throw new \InvalidArgumentException('Unknown parser kind');
+            }
+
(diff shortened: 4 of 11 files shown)


GrahamCampbell:
:-1:

GrahamCampbell:
We should retain compatibility with both 1.x and 2.x.

composer.json
```diff
@@ -13,10 +13,10 @@
         }
     ],
     "require": {
-        "php": ">=5.3.9",
+        "php": ">=5.4",
```
GrahamCampbell: Should be reverted.

src/Psy/CodeCleaner/ValidFunctionNamePass.php
```diff
@@ -58,7 +59,7 @@ public function leaveNode(Node $node)
         if ($node instanceof FuncCall) {
             // if function name is an expression or a variable, give it a pass for now.
             $name = $node->name;
-            if (!$name instanceof Expression && !$name instanceof Variable) {
+            if (!$name instanceof Expr && !$name instanceof Variable) {
```
GrahamCampbell: ?

src/Psy/Command/ParseCommand.php
```diff
@@ -126,7 +127,9 @@ private function parse($code)
     private function getParser()
     {
         if (!isset($this->parser)) {
-            $this->parser = new Parser(new Lexer());
+            $this->parser = (new ParserFactory())->create(
+                version_compare(PHP_VERSION, '7.0', '>=') ? ParserFactory::ONLY_PHP7 : ParserFactory::ONLY_PHP5
```
GrahamCampbell: ?

GrahamCampbell: just use the prefer php 7 one, and be done with it surely?

src/Psy/Command/ParseCommand.php
```diff
@@ -14,6 +14,7 @@
 use PhpParser\Lexer;
```
GrahamCampbell: this is now an unused import

GrahamCampbell:
Something like this would be fine: https://github.com/jeremeamia/super_closure/pull/70/files.

src/Psy/CodeCleaner/ValidFunctionNamePass.php
alexeyshockov: I think it's a bug, because PHP parser doesn't have `Expression` type. It has only `Expr`.

alexeyshockov:
I agree, that support for both versions of PHP parser is better.

src/Psy/Command/ParseCommand.php
alexeyshockov: In the `CodeCleaner` strict options should be used, because parsed code will be evaluated on the same interpreter (so with `PREFER_PHP7` invalid code can be passed to 5.x interpreter).

I don't have strong opinion about `ParseCommand`. I think the best way will be to add additional console option for PHP version. What do you think?

src/Psy/CodeCleaner/ValidFunctionNamePass.php
GrahamCampbell: That should be fixed in a separate PR then.

src/Psy/Command/ParseCommand.php
```diff
@@ -126,7 +126,9 @@ private function parse($code)
     private function getParser()
     {
         if (!isset($this->parser)) {
-            $this->parser = new Parser(new Lexer());
+            $this->parser = (new ParserFactory())->create(
```
GrahamCampbell: this class doesn't exist on 1.x

GrahamCampbell: you need to update this

GrahamCampbell: you can't just update the version constraint to allow 1.x again, but leave the code here that only works on 2.x

alexeyshockov: I already understood, don't work as CI ;)

src/Psy/CodeCleaner/ValidFunctionNamePass.php
alexeyshockov: Agree. Moved to a separated PR (#252). Thanks!

src/Psy/Command/ParseCommand.php
alexeyshockov: Now it's used again (after reverting support for PHP parser 1.x).

alexeyshockov:
So, tests pass. Only one question about proper option to PHP parser (`PREFER_*` or `ONLY_*`) is opened now, as I can see.

src/Psy/CodeCleaner.php
```diff
@@ -53,7 +54,17 @@ class CodeCleaner
      */
     public function __construct(Parser $parser = null, Printer $printer = null, NodeTraverser $traverser = null)
     {
-        $this->parser    = $parser    ?: new Parser(new Lexer());
+        if (class_exists('PhpParser\ParserFactory')) {
```
GrahamCampbell: this code is duplicated - we need to find a way to avoid that

GrahamCampbell:
Maybe the other class alias fixes should be moved to another PR too?

GrahamCampbell:
Maybe just make them two commits and keep them both in here, then just rebase if the other PRs are merged?

alexeyshockov:
Splited to two commits.

bobthecow:
Thanks for getting this going!

Let's make our own parser factory of some sort so we don't duplicate the parser creation logic. It should probably have an optional "support php 7" flag, which defaults to "is this currently php 7?" but could be passed as an arg from the parse command.

alexeyshockov:
Good, I added own factory. It just forwards original PHP parser factory constants, because `parse` command doesn't abstract user over the library, it's only an interface.

src/Psy/CodeCleaner.php
```diff
@@ -53,7 +52,9 @@ class CodeCleaner
      */
     public function __construct(Parser $parser = null, Printer $printer = null, NodeTraverser $traverser = null)
     {
-        $this->parser    = $parser    ?: new Parser(new Lexer());
+        $parserFactory = new ParserFactory();
```
GrahamCampbell: hmmm, only create the parser factory if we need it surely?

GrahamCampbell: ie, `$this->parser    = $parser    ?: (new ParserFactory())->createParser();`

alexeyshockov: Sad, but not working with PHP 5.3 :(

GrahamCampbell: we should add an if statement then

bobthecow: But this works on 5.3 :)

```php
        if ($parser === null) {
            $parserFactory = new ParserFactory();
            $parser = $parserFactory->createParser();
        }

        $this->parser    = $parser;
```

GrahamCampbell: > But this works on 5.3 :)

Yeh. :P

src/Psy/Command/ParseCommand.php
```diff
@@ -59,12 +68,23 @@ public function setPresenter(Presenter $presenter)
...
     {
+        $definition = array(
+            new InputArgument('code', InputArgument::REQUIRED, 'PHP code to parse.'),
+            new InputOption('depth', '', InputOption::VALUE_REQUIRED, 'Depth to parse', 10),
+        );
+        if ($this->parserFactory->hasKindsSupport()) {
+            $definition[] = new InputOption(
+                'parser-kind',
+                '',
+                InputOption::VALUE_OPTIONAL,
```
bobthecow: This should be `VALUE_REQUIRED`.

src/Psy/Command/ParseCommand.php
```diff
@@ -59,12 +68,23 @@ public function setPresenter(Presenter $presenter)
      */
     protected function configure()
     {
+        $definition = array(
+            new InputArgument('code', InputArgument::REQUIRED, 'PHP code to parse.'),
+            new InputOption('depth', '', InputOption::VALUE_REQUIRED, 'Depth to parse', 10),
+        );
+        if ($this->parserFactory->hasKindsSupport()) {
+            $definition[] = new InputOption(
+                'parser-kind',
```
bobthecow: I'm not sure I love `parser-kind` as an option name. I don't have a better suggestion right now, other than to drop the `parser-` prefix, since that's kind of implied by the command :)

src/Psy/ParserFactory.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getDefaultKind()
+    {
+        if (self::hasKindsSupport()) {
+            return version_compare(PHP_VERSION, '7.0', '>=') ? self::ONLY_PHP7 : self::ONLY_PHP5;
+        } else {
```
bobthecow: the `else return` is unnecessary.

src/Psy/ParserFactory.php
```diff
@@ -0,0 +1,88 @@
...
+            $kind = $kind ?: self::getDefaultKind();
+
+            if (!in_array($kind, self::getPossibleKinds())) {
+                throw new \InvalidArgumentException('Unknown parser kind');
+            }
+
+            $parser = $originalFactory->create(constant('PhpParser\ParserFactory::' . $kind));
+        } else {
+            if ($kind !== null) {
+                throw new \InvalidArgumentException('You must install second version of PHP parser to specify kind.');
```
bobthecow: `Install PHP Parser v2.x to specify parser kind` or something like that?

src/Psy/ParserFactory.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Available kinds for second version of PHP parser.
+     *
+     * @return array
+     */
+    private function getPossibleKinds()
+    {
+        $factory = new ReflectionClass(__CLASS__);
+
+        return array_values($factory->getConstants());
```
bobthecow: Why don't we add a private static property rather than using reflection for something that will almost never change?

src/Psy/Command/ParseCommand.php
```diff
@@ -59,12 +68,23 @@ public function setPresenter(Presenter $presenter)
      */
     protected function configure()
     {
+        $definition = array(
+            new InputArgument('code', InputArgument::REQUIRED, 'PHP code to parse.'),
+            new InputOption('depth', '', InputOption::VALUE_REQUIRED, 'Depth to parse', 10),
+        );
```
bobthecow: Add a newline here for readability.

src/Psy/Command/ParseCommand.php
```diff
@@ -59,12 +68,23 @@ public function setPresenter(Presenter $presenter)
...
+        if ($this->parserFactory->hasKindsSupport()) {
+            $definition[] = new InputOption(
+                'parser-kind',
+                '',
+                InputOption::VALUE_OPTIONAL,
+                "One PhpParser\\ParserFactory constants (default is based on current interpreter's version)",
+                $this->parserFactory->getDefaultKind()
```
bobthecow: I'd like to enumerate the possible parser kinds, probably in the `<<<HELP` block below? Also wouldn't hurt to add an example while we're at it :)

src/Psy/ParserFactory.php
alexeyshockov: Agree. Fixed.

src/Psy/ParserFactory.php
alexeyshockov: Fixed.

src/Psy/Command/ParseCommand.php
alexeyshockov: I added all possibilities to the option description. I think, that's enough :)

src/Psy/Command/ParseCommand.php
alexeyshockov: Fixed.

src/Psy/Command/ParseCommand.php
alexeyshockov: Agree with simple `kind`. Fixed.

src/Psy/Command/ParseCommand.php
alexeyshockov: Fixed.

src/Psy/CodeCleaner.php
alexeyshockov: Agree, fixed to `if`. Hope we will see PHP version increment in the requirements soon...

src/Psy/ParserFactory.php
alexeyshockov: I don't like to duplicate things this way (yet another string with the same constant names). But it's not so important for me, so if you think differently...

src/Psy/ParserFactory.php
```diff
@@ -0,0 +1,86 @@
...
+     * @return array
+     */
+    public static function getPossibleKinds()
+    {
+        $factory = new ReflectionClass(__CLASS__);
+
+        return array_values($factory->getConstants());
+    }
+
+    /**
```
GrahamCampbell: missing short description

src/Psy/ParserFactory.php
```diff
@@ -0,0 +1,86 @@
...
+     * @return bool
+     */
+    public function hasKindsSupport()
+    {
+        return class_exists('PhpParser\ParserFactory');
+    }
+
+    /**
```
GrahamCampbell: missing short description

src/Psy/ParserFactory.php
```diff
@@ -0,0 +1,86 @@
...
+     * @return string|null
+     */
+    public function getDefaultKind()
+    {
+        if ($this->hasKindsSupport()) {
+            return version_compare(PHP_VERSION, '7.0', '>=') ? static::ONLY_PHP7 : static::ONLY_PHP5;
+        }
+    }
+
+    /**
```
GrahamCampbell: missing short description

src/Psy/Command/ParseCommand.php
```diff
@@ -28,7 +28,16 @@
 class ParseCommand extends Command implements PresenterAware
 {
     private $presenter;
-    private $parser;
+    private $parserFactory;
+    private $parsers;
+
+    public function __construct($name = null)
```
GrahamCampbell: some phpdoc would be nice here so we know what name actually is

GrahamCampbell:
Asside from a few more minor things I've just commented, this is looking pretty good now. :+1:

src/Psy/CodeCleaner.php
bobthecow: > Hope we will see PHP version increment in the requirements soon...

Probably not, at least until it becomes too arduous to support. PsySH is useful even for people whose legacy support needs preclude them from using PHP 5.4+ :)

src/Psy/ParserFactory.php
bobthecow: I'd prefer changing it. I don't like reflection at runtime any more than we have to.

And yeah, I know, I'm the creator of a REPL that does an obscene amount of runtime reflection :)

src/Psy/Command/ParseCommand.php
bobthecow: just use an inheritDoc tag :)

bobthecow:
:+1:

Fill in the missing docs that @GrahamCampbell pointed out, and change that one bit of reflection, and we're good to merge. Thanks!

bobthecow:
@alexeyshockov around? i've got a bit of time right now, and won't have a chance to look at this again for a week or so, so i wanted to merge soon :)

alexeyshockov:
I'm going to send updates later today, at night. I hope it will not be late!

On Saturday, November 7, 2015, Justin Hileman <notifications@github.com>
wrote:

> @alexeyshockov <https://github.com/alexeyshockov> around? i've got a bit
> of time right now, and won't have a chance to look at this again for a week
> or so, so i wanted to merge soon :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bobthecow/psysh/pull/251#issuecomment-154718813>.
>


-- 
Alexey Shockov

tel:+79652121992
skype:alexeyshockov
mailto:alexey@shockov.com
http://alexey.shockov.com/

src/Psy/ParserFactory.php
alexeyshockov: Fixed.

src/Psy/ParserFactory.php
alexeyshockov: Fixed.

alexeyshockov:
Fixed all issues, please review.

bobthecow:
Looking good. Thank you!

Issue #233, opened 2015-08-26
Can't find the documentation file for PHP


digfish:
On a Laravel 5 project under Windows 10 Home (64-bit,portuguese), psysh is not able to find the documentation database file (php_manual.sqlite) under the standard location C:\Users\[username]\AppData\Roaming\PsySh .
 Usually I start PsySh under laravel with

`php artisan tinker` 

under the root directory of the project.
I tried to use a configuration file for PsySh like the one it is quoted in the github page and call it at the start of execution of psysh with:

`php artisan tinker -vvv ./psysh-config.php`

but that does not work, either.

bobthecow:
Sorry about the delayed response.

Is your filesystem case sensitive? It should be `PsySH` not `PsySh`, so that might be it.

Otherwise, run `Psy\info()` in psysh and let me know what it says?

jaden:
I'm running psysh globally with the same problem of not finding the documentation DB file.

It's located here: C:\Users\dan\AppData\Roaming\PsySH\php_manual.sqlite 

This is the output of Psy\info()
```
>>> Psy\info()
=> [
     "PsySH version" => "v0.5.2",
     "PHP version" => "5.6.13",
     "default includes" => [],
     "require semicolons" => false,
     "config file" => [
       "default config file" => null,
       "PSYSH_CONFIG env" => false,
     ],
     "pcntl" => [
       "pcntl available" => false,
       "posix available" => false,
     ],
     "readline" => [
       "readline available" => false,
     ],
     "history" => [
       "history file" => "C:\Users\dan\.config/psysh_history",
       "history size" => null,
       "erase duplicates" => null,
     ],
     "docs" => [
       "manual db file" => null,
       "sqlite available" => true,
     ],
     "autocomplete" => [
       "tab completion enabled" => false,
       "custom matchers" => [],
     ],
   ]
```

digfish:
Sorry for answering just now, here is the output of Psy\info() on my system:
```
=> [
     "PsySH version" => "v0.6.0-dev",
     "PHP version" => "5.6.11",
     "default includes" => [],
     "require semicolons" => false,
     "config file" => [
       "default config file" => null,
       "PSYSH_CONFIG env" => false,
     ],
     "pcntl" => [
       "pcntl available" => false,
       "posix available" => false,
     ],
     "readline" => [
       "readline available" => false,
     ],
     "history" => [
       "history file" => "C:\Users\sam\.config/psysh_history",
       "history size" => null,
[...]

PR #253, opened 2015-11-03
Documentation fix for issue #233.

jaden:
The documentation was pointing to C:\Users\{USER}\AppData\Roaming\PsySH
but PsySH tests the following directories:

C:\Users\{USER}/.psysh/php_manual.sqlite
C:\Users\{USER}\.local\share/psysh/php_manual.sqlite
/usr/local/share/psysh/php_manual.sqlite
/usr/share/psysh/php_manual.sqlite

Since the first directory above is just as good as AppData, I updated the
README to suggest using it on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ return array(
 
 ## Downloading the manual
 
-The PsySH `doc` command is great for documenting source code, but you'll need a little something extra for PHP core documentation. Download one of the following PHP Manual files and drop it in `~/.local/share/psysh/` (or `C:\Users\{USER}\AppData\Roaming\PsySH` on Windows):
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Should be targetted at the develop branch rather than the master.

bobthecow:
The first directory isn't just as good as AppData. Just like XDG is the "right" place to put these files on *nix, AppData is the "right" place to put them on Windows. The fact that the existence of a HOME environment variable throws off the AppData support is a bug which should be fixed :)

jaden:
@GrahamCampbell Thanks for the heads up.

@bobthecow I can see that. On the other hand, I currently have 49 dot directories in my HOME directory :)

bobthecow:
> On the other hand, I currently have 49 dot directories in my HOME directory :)

PsySH _tries_ to be better than those 48 other libraries, though ;)

jaden:
> PsySH tries to be better than those 48 other libraries, though ;)

I admire your principled adherence to Windows standards. :) 
And after further review, many of the apps using HOME are nix-first and run on Windows, but were not Windows native.

bobthecow:
:)

PR #256, opened 2015-11-07
Switched to PSR4 for autoloading

GrahamCampbell:
PSR0 is deprecated, and we can move to PSR4 without moving any files about.


GrahamCampbell:
Does this look ok @bobthecow?

bobthecow:
Yep, merged!

PR #257, opened 2015-11-07
Additional CS Fixes

GrahamCampbell:


.php_cs
```diff
@@ -21,7 +34,10 @@ $config = Config::create()
     ->setUsingLinter(false);
 
 $finder = $config->getFinder()
-    ->in('src')
-    ->in('test');
+    ->in(__DIR__)
```
bobthecow: is it smart enough to exclude vendor if we run it on `__DIR__`?

bin/psysh
```diff
@@ -82,7 +82,6 @@ call_user_func(function() {
 // We didn't find an autoloader for a local version, so use the autoloader that
 // came with this script.
 if (!class_exists('\Psy\Shell')) {
-/* <<< */
```
bobthecow: These comments are necessary, they're how the compiler knows where to put the phar autoloader code.

GrahamCampbell: php-cs-fixer was moving these comments about, so i removed them - i'll see what i can do here

.php_cs
GrahamCampbell: yes, it is :)

GrahamCampbell: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Finder/DefaultFinder.php

GrahamCampbell:
How about that?

.php_cs
bobthecow: :+1:

bobthecow:
How did you stop it from moving the comments around?

GrahamCampbell:
I just stopped fixing the file. We weren't fixing it before anyway.

GrahamCampbell:
https://github.com/GrahamForks/psysh/commit/b3af2a5227544779faeb22ce022441d358c2be52

bobthecow:
Heh. I suppose that works :)

.php_cs
bobthecow: This fails if you've ever done a `bin/build` because there's a second vendor directory just for phar builds called `build-vendor`, and it checks that. Mind adding that as an exclusion?

(And yeah, I really should test phar builds on CI)

GrahamCampbell: Sure. :)

GrahamCampbell:
All done @bobthecow. :)

bobthecow:
Thanks!

PR #258, opened 2015-11-08
Autoload the tests in composer.json

GrahamCampbell:
I've also removed the custom bootstrap file since it's pretty common to just try to require the file composer gives, and most people can recognise the error php will give if it fails to require that file, and then go and run composer if they didn't run it already.


bobthecow:
:+1:

PR #261, opened 2015-11-12
fix incorrect parameter

mix5003:


bobthecow:
Craaap. That's what I get for mocking that entire code path while testing it, rather than spinning up a Windows machine :(

bobthecow:
Thanks for the fix. I'll get a release cut tonight.

bobthecow:
Thanks again. This has been merged and released as a hotfix in v0.6.1.

PR #263, opened 2015-11-16
Only default commands get autocompletion

damiankloip:
Now that we have drush commands inside PsySH in drush (which is awesome!) one small problem is that the autocompletion does not work for any of these commands.

I looked into it further; Only the default commands will get this autocompletion as tab completion is initialized in the `Shell` constructor. So anything using `addCommands`, like we are, is too late. The `CommandsMatcher` instance has already been created with `$this->all()` - which at the time is only the default commands.

This just moves the tab completion code into an `initializeTabCompletion` method, which then gets called in `Shell::run` instead. This means everything added before the shell is run will be included.

Not sure if this is the best way to fix this problem? Certainly seems like the simplest and least invasive approach maybe. Also confirmed this fixes the completion of added commands. Thoughts?

### DIFF
--- a/src/Psy/Shell.php
+++ b/src/Psy/Shell.php
@@ -78,20 +78,6 @@ class Shell extends Application
         $this->readline = $this->config->getReadline();
 
         parent::__construct('Psy Shell', self::VERSION);
-
-        $this->config->setShell($this);
-        // auto completer needs shell to be linked to configuration because of the context aware matchers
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Please send to the develop branch.

damiankloip:
Thanks, closing in favour of https://github.com/bobthecow/psysh/pull/266

PR #264, opened 2015-11-16
Make 'exit;' behave the same as the exit command

damiankloip:
Add an ExitPass so `exit;` behaves the same as the exit command.

This should be helping us soon in the drush psysh integration (we have problems with the database connection being closed before the drush command has finished, when the shell exits).

### DIFF
--- a/src/Psy/CodeCleaner/ExitPass.php
+++ b/src/Psy/CodeCleaner/ExitPass.php
@@ -12,12 +12,11 @@
 namespace Psy\CodeCleaner;
 
 use PhpParser\Node;
-use Psy\Exception\BreakException;
 use PhpParser\Node\Expr\Exit_;
+use Psy\Exception\BreakException;
 
 class ExitPass extends CodeCleanerPass
 {
-
     /**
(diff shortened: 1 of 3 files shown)


src/Psy/CodeCleaner/ExitPass.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Throws a PsySH BreakException instead on exit().
+     *
+     * @throws \Psy\Exception\BreakException if the node is an exit node.
+     *
+     * @param \PhpParser\Node $node
+     */
+    public function enterNode(Node $node) {
+        if ($node instanceof Exit_) {
+            throw new BreakException('Goodbye.');
```
dawehner: Could we convert it to a command instead first? I guess this is not possible?

dawehner:
Really nice!

GrahamCampbell:
Please send to the develop branch.

damiankloip:
Thanks, closing in favour of https://github.com/bobthecow/psysh/pull/265

PR #265, opened 2015-11-16
Make 'exit;' behave the same as the exit command

damiankloip:
Add an ExitPass so exit; behaves the same as the exit command.

This should be helping us soon in the drush psysh integration (we have problems with the database connection being closed before the drush command has finished, when the shell exits).

### DIFF
--- a/test/Psy/Test/CodeCleaner/ImplicitReturnPassTest.php
+++ b/test/Psy/Test/CodeCleaner/ImplicitReturnPassTest.php
@@ -33,6 +33,7 @@ class ImplicitReturnPassTest extends CodeCleanerTestCase
         return array(
             array('4',     'return 4;'),
             array('foo()', 'return foo();'),
+            array('exit()', 'die;'),
         );
     }
 }
--- a/src/Psy/CodeCleaner.php
+++ b/src/Psy/CodeCleaner.php
@@ -18,6 +18,7 @@ use Psy\CodeCleaner\AbstractClassPass;
 use Psy\CodeCleaner\AssignThisVariablePass;
 use Psy\CodeCleaner\CalledClassPass;
 use Psy\CodeCleaner\CallTimePassByReferencePass;
+use Psy\CodeCleaner\ExitPass;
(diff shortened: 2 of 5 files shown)


bobthecow:
Nice!

What happens if you do this:

```php
if (false) {
  exit;
} else {
  echo "do something cool!";
}
```

:)

GrahamCampbell:
What's going on with StyleCI here? Does this need rebasing?

bobthecow:
Master is failing on StyleCI, so since this was originally created against master it might be failing as well?

@damiankloip Do you mind rebasing against the dev branch?

damiankloip:
Sorry, forgot to merge develop from upstream.

damiankloip:
@bobthecow Youch! Fair point. Any ideas about the best way to not bail out at the first sniff of exit? :)

bobthecow:
You could probably rewrite the AST to throw an exception on execution, rather than throwing an exception on parse?

damiankloip:
I did try something on leaveNode() but this seems to get out of hand fast..

```php
return new Node\Stmt\Throw_(new Node\Expr\New_(new Node\Name('Psy\Exception\BreakException'), [new Node\Arg(new Node\Scalar\String_('Goodbye.'))]));
```

damiankloip:
@bobthecow Yeah, seems the `ImplicitReturnPass` was an issue. Seems to work much better now. Please give it a try.

damiankloip:
Just updated the tests for the change to using `leaveNode` too.

test/Psy/Test/CodeCleaner/ImplicitReturnPassTest.php
```diff
@@ -33,6 +33,7 @@ public function implicitReturns()
         return array(
             array('4',     'return 4;'),
             array('foo()', 'return foo();'),
+            array('exit()', 'die;'),
```
damiankloip: Thought it makes sense to also add a new item to the data provider to check that no implicit return is added now.

bobthecow:
Looks good to me. Thanks!

Issue #78, opened 2014-07-02
No colors or ">>>" prompt when using in phpunit


wulftone:
I can put the `\Psy\Shell::debug(get_defined_vars());` into my phpunit tests and it drops into a shell just fine, but the colors do not work, and the `>>>` prompt doesn't appear--it's just blank.  It doesn't seem to be interpreting any arrow key codes either.

It still works well enough, but these features are pretty nice to have, and they DO work when I just run `psysh` on its own.  Any ideas?

kenjis:
I have colored output but no `>>>` at first. But once an error is outputted, `>>>` appears.

~~~
PHPUnit 4.8.6 by Sebastian Bergmann and contributors.

Psy Shell v0.6.0-dev (PHP 5.5.26 — cli) by Justin Hileman
ls
Variables: $mock, $mock2, $this
ls
Variables: $mock, $mock2, $this
show
RuntimeException with message 'Not enough arguments.'
>>> ls
Variables: $mock, $mock2, $this
>>> 
~~~

toddmazierski:
Hi, all! I had trouble getting ANSI colors to appear in PHPUnit when debugging with PsySH, too. When running PHPUnit (a fairly old version—4.1.5), the following value:

```php
@posix_isatty(STDOUT)
```

…returns `false`, so the following two dependencies automatically disable colors:

* [`StreamOutput` in Symfony](https://github.com/symfony/symfony/blob/2dbb75aef49c39ac680a31f44514f94c29de681d/src/Symfony/Component/Console/Output/StreamOutput.php#L99)
* [`ConsoleColor` in PHP Console Color](https://github.com/JakubOnderka/PHP-Console-Color/blob/6855b75259010f67e8425aff1259c6e3f63750b7/src/JakubOnderka/PhpConsoleColor/ConsoleColor.php#L207)

Both libraries _do_ support overriding this auto-detection mechanism (`$decorated` in Symfony, `setForceStyle()` in PHP Console Color), so perhaps Psysh could support a configuration flag itself, and pass it along.

Hope that's useful!

bobthecow:
This should be solved by `colorMode` config, and `—color` and `—no-color` command line flags :)

mkolom:
`colorMode` setting worked for me as for colors but what about prompt and, more importantly, arrow codes?
Is it possible to use arrows from within PHPUnit? Outside it works well.
Readline is available and enabled.

igorsantos07:
[...]

PR #271, opened 2015-12-19
Feature proposal: add configurable `forceColor` option

toddmazierski:
When `true`, forces output to contain ANSI colors, even in non-interactive (non-tty) shells, where it may not be supported.

This may be a solution for the participants in #78. Thanks for looking!

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,10 @@ return array(
     //
     // This will default to true in a future release, but is false for now.
     'warnOnMultipleConfigs' => true,
+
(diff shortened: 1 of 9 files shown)


GrahamCampbell:
Please send PRs to the develop branch.

toddmazierski:
Will do, @GrahamCampbell, as soon as I correct these failed checks!

toddmazierski:
Moved to #272.

PR #272, opened 2015-12-19
Feature proposal: add configurable `forceColor` option

toddmazierski:
When `true`, forces output to contain ANSI colors, even in non-interactive (non-tty) shells, where it may not be supported.

This may be a solution for the participants in #78. Thanks for looking!


src/Psy/Command/WhereamiCommand.php
```diff
@@ -108,8 +110,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $contents = file_get_contents($info['file']);
         $output->page($highlighter->getCodeSnippet($contents, $info['line'], $num, $num), ShellOutput::OUTPUT_RAW);
     }
+
+    /**
+     * Sets whether or not to force colors in output.
+     *
+     * @param bool $forceColor
+     */
+    public function setForceColor($forceColor)
```
bobthecow: I would personally use a constructor argument for this, rather than getters and setters. That simplifies things here as well as `Shell::getDefaultCommands` :)

bobthecow:
:+1: 

Does it make sense to add support for invoking force color via a command line flag as well?

```
$ psysh --force-color ...
```

src/Psy/Command/WhereamiCommand.php
toddmazierski: Ah! I was wondering about that. Will do. :+1:

toddmazierski:
> Does it make sense to add support for invoking force color via a command line flag as well?

It does to me! I'll make that change. :+1:

toddmazierski:
Ready for re-review, @bobthecow! Thanks for the feedback. :smile: 

I'll squash everything into a single commit if it's a "go" to merge.

bobthecow:
looks good to me.

The one other thing to consider is whether to have a three-state color option, i.e. true/false/auto. IIRC some windows users run into trouble with colors with some combination of cygwin or third party terminals, or something. If we go that route, flags could be `--color` and `--no-color`, corresponding to `'forceColor' => true` and `'forceColor' => false`. If not set at all, it would be auto as it currently does by default.

toddmazierski:
> The one other thing to consider is whether to have a three-state color option, i.e. true/false/auto.

Both colorization libraries PsySH uses _are_ taking Windows terminals into account in their auto-detection functions:

* [`StreamOutput` in Symfony](https://github.com/symfony/symfony/blob/2dbb75aef49c39ac680a31f44514f94c29de681d/src/Symfony/Component/Console/Output/StreamOutput.php#L83-L100)
* [`ConsoleColor` in PHP Console Color](https://github.com/JakubOnderka/PHP-Console-Color/blob/6855b75259010f67e8425aff1259c6e3f63750b7/src/JakubOnderka/PhpConsoleColor/ConsoleColor.php#L198-L208)

Because of this, I don't think PsySH needs to be concerned about a third `--no-color` configuration state.

That said, I'm curious to hear your thoughts, and I'm happy to help if it's decided that's the way to go. :smile:

bobthecow:
> Both colorization libraries PsySH uses are taking Windows terminals into account in their auto-detection functions:

Right. I seem to recall something about that sort of auto detection failing in some circumstances, which is why most Symfony console based apps (e.g. `composer`) have `--ansi` and `--no-ansi` flags for forcing either direction.

That said, I don't have a strong opinion either way :)

toddmazierski:
> …most Symfony console based apps (e.g. `composer`) have `--ansi` and `--no-ansi` flags for forcing either direction.

Hmm…consistency _is_ important. And having to deprecate & rename a configuration flag just because we didn't take into account all possible states sounds annoying.

I'm convinced. :smile:  If you'll allow me a few more days, I'll make that modification. Thanks for discussing, @bobthecow!

bobthecow:
Sounds good. Thanks @toddmazierski :)

bobthecow:
@toddmazierski Any update on this? Thanks!

toddmazierski:
My apologies for the delay, @bobthecow! I'll wrap this up this weekend.

bobthecow:
No worries. Thanks

toddmazierski:
Moving to #279!

bobthecow:
:+1:

PR #279, opened 2016-01-18
Feature proposal: color modes to force/disable colors in output

toddmazierski:
Moved from #272!

Changes in detail:

* Modify `ShowCommand`, `WhereamiCommand`, `Shell`, and `CodeFormatter` to be aware of color modes

* ~~Add `ColorMode` configuration object to represent available and current color modes~~

* Add `ConsoleColorFactory` to manage configuring `ConsoleColor` instances with the current color mode

* Add `colorMode` option to configuration

* Add `--color` and `--no-color` command line flags

* Update README

* Add tests


src/Psy/functions.php
```diff
@@ -155,6 +157,16 @@ function bin()
                 $config['configFile'] = $configFile;
             }
 
+            // Handle --color
+            if ($input->getOption('color')) {
+                $config['colorMode'] = ColorMode::getForced();
+            }
+
+            // Handle --no-color
+            if ($input->getOption('no-color')) {
```
bobthecow: Let's throw an error if both --color and --no-color are specified.

src/Psy/Formatter/CodeFormatter.php
```diff
@@ -38,8 +42,7 @@ public static function format(\Reflector $reflector)
             $start = $reflector->getStartLine();
             $end   = $reflector->getEndLine() - $start;
 
-            $colors = new ConsoleColor();
-            $colors->addTheme('line_number', array('blue'));
+            $colors = (new ConsoleColorFactory($colorMode))->getConsoleColor();
```
bobthecow: Isn't this a 5.4+ thing? PsySH still supports 5.3 :)

src/Psy/ColorMode.php
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Psy;
+
+/**
+ * A configuration object that represents one of 3 available color modes:
+ *   1. "auto" (the default): if color support is detected, output contains colors
+ *   2. "forced": force colors in output
+ *   3. "disabled": disable colors in output
+ */
+class ColorMode
```
bobthecow: TBH I don't see the advantage of using a class for this over just using three constants and checking equality in services that consume it. Convince me?

src/Psy/Formatter/CodeFormatter.php
toddmazierski: Ah, I didn't know that! These [5.3 test failures](https://travis-ci.org/bobthecow/psysh/jobs/103157241) confirm it.

src/Psy/functions.php
toddmazierski: :+1:

src/Psy/ColorMode.php
toddmazierski: > Convince me?

I'm happy to try! If unsuccessful, constants and equality it is. :+1: A few design reasons to consider an alternative, though:

### Type hinting

Because color mode is a class, type hinting can be used to help prevent bugs (and aid in debugging existing ones).

#### Example: class and functions

If an invalid value is assigned, a clear error happens immediately:

```php
(new \Psy\Configuration)->setColorMode('oops');
```
```
PHP error:  Argument 1 passed to Psy\Configuration::setColorMode() must be an instance of Psy\ColorMode, string given, called in...
```

#### Example: constants and equality

If an invalid value is assigned, nothing happens immediately:

```php
(new \Psy\Configuration)->setColorMode('oops'); // => null
```

However, a less clear error might happen later on (ex. `call on a non-object`).

### Silent failures

Equality can lead to silent failures, since nearly anything can be compared to anything.

#### Example: class and functions

For example, imagine that the `colorMode` has not been assigned due to an initialization bug, so it's `null`:

```php
$config->colorMode(); // => null

if ($config->colorMode()->isAuto()) {
    // ...
}
```

A fairly clear error happens immediately:

```
PHP Fatal error:  Call to a member function isAuto() on a non-object...
```

#### Example: constants and equality

However, imagine the same scenario as above, but with constants and equality:

```php
$config->colorMode(); // => null

if ($config->colorMode() === \Psy::COLOR_MODE_AUTO) {
    // ...
}
```

No error occurs. However, a less clear error might happen later on.

### Skipped use of constants

Developers are sometimes tempted (myself included) to simply skip using the constants (ex. to keep their code terse). This can lead to unintended bugs:

```php
if ($config->colorMode() === 'automatic') {
    // (will never execute since "automatic" is not a valid setting)
}
```

### Unintented operators

Since equality is not a concrete, public interface, developers may be tempted to use other operators the author didn't intend. For example, `==` instead of `===`, or worse:

```php
if (preg_match('/auto/', $config->colorMode())) {
    // ...
}
```

There are a couple of other reasons (difficult to mock in tests, bound to static behavior), but I think these are the most complelling! Thanks for considering it!

bobthecow: It feels like all of these are twisting ourselves into contortions to work around PHP's lack of enums and the ability to typehint against them :)

Remember that these are personal configuration options, not mission critical framework integrations. If something goes wrong, the user just has the wrong color preferences.

In keeping with the Robustness Principle, operators don't really matter. Sure, you _should_ be using the constants. But if you pass something in that `==` the value of the constant, that should be good enough for PsySH. The same way we don't throw an error if you pass something that `==` false to one of the boolean configs, we just silently convert it into what we were expecting.

Now, if you pass something that isn't `==` one of the available options, it's totally reasonable to validate it when setting and throw an exception. But we could do that with one line of code, vs the 100 line class approach.

IFF this were a strongly typed language, with expressive typehinting, and enums, and all sorts of other things, then we could use those language features to do this "right". But we still wouldn't use classes for it :P

src/Psy/ColorMode.php
```diff
@@ -0,0 +1,102 @@
...
+ * (c) 2012-2015 Justin Hileman
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy;
+
+/**
+ * A configuration object that represents one of 3 available color modes:
```
GrahamCampbell: CS checks failing because this is not a valid short description. Provide one first, then leave a newline, and add this description.

src/Psy/ColorMode.php
toddmazierski: > It feels like all of these are twisting ourselves into contortions to work around PHP's lack of enums and the ability to typehint against them :)

Ha! Now that may very well be the case. :smile:

That said, I very much appreciate you entertaining my arguments, @bobthecow! Reworking the code now to use constants and equality.

src/Psy/ColorMode.php
toddmazierski: :+1: This class has been removed entirely, so the checks should come back clean now.

src/Psy/ConsoleColorFactory.php
```diff
@@ -0,0 +1,76 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy;
+
+use JakubOnderka\PhpConsoleColor\ConsoleColor;
+use JakubOnderka\PhpConsoleHighlighter\Highlighter;
+
+class ConsoleColorFactory
```
bobthecow: Mind adding a docblock to this class?

bobthecow:
Looking good. Add a comment to that one class and squash everything?

toddmazierski:
DocBlock added and commits squashed, @bobthecow! Thanks for all your help.

bobthecow:
Great. Thank _you_!

Issue #297, opened 2016-04-02
string assignment causes PHP Fatal error


condy0919:
the output is 

Psy Shell v0.7.2 (PHP 7.0.5 — cli) by Justin Hileman

> > > $text = "hello"
> > > PHP Fatal error:  Uncaught Error: Call to undefined function Symfony\Component\VarDumper\Dumper\iconv_strlen() in phar:///usr/bin/psysh/build-vendor/symfony/var-dumper/Dumper/CliDumper.php:172
> > > Stack trace:
> > > # 0 phar:///usr/bin/psysh/build-vendor/symfony/var-dumper/Cloner/Data.php(179): Symfony\Component\VarDumper\Dumper\CliDumper->dumpString(Object(Symfony\Component\VarDumper\Cloner\Cursor), 'hello', false, 0)
> > > # 1 phar:///usr/bin/psysh/build-vendor/symfony/var-dumper/Cloner/Data.php(91): Symfony\Component\VarDumper\Cloner\Data->dumpItem(Object(Psy\VarDumper\Dumper), Object(Symfony\Component\VarDumper\Cloner\Cursor), Array, 'hello')
> > > # 2 phar:///usr/bin/psysh/build-vendor/symfony/var-dumper/Dumper/AbstractDumper.php(121): Symfony\Component\VarDumper\Cloner\Data->dump(Object(Psy\VarDumper\Dumper))
> > > # 3 phar:///usr/bin/psysh/src/Psy/VarDumper/Presenter.php(119): Symfony\Component\VarDumper\Dumper\AbstractDumper->dump(Object(Symfony\Component\VarDumper\Cloner\Data), Object(Closure))
> > > # 4 phar:///usr/bin/psysh/src/Psy/Shell.php(741): Psy\Va in phar:///usr/bin/psysh/build-vendor/symfony/var-dumper/Dumper/CliDumper.php on line 172

Fatal error: Uncaught Error: Call to undefined function Symfony\Component\VarDumper\Dumper\iconv_strlen() in phar:///usr/bin/psysh/build-vendor/symfony/var-dumper/Dumper/CliDumper.php on line 172

Error: Call to undefined function Symfony\Component\VarDumper\Dumper\iconv_strlen() in phar:///usr/bin/psysh/build-vendor/symfony/var-dumper/Dumper/CliDumper.php on line 172

Call Stack:
    0.0010     380976   1. {main}() /usr/bin/psysh:0
[...]

PR #300, opened 2016-04-12
Add HoaConsole readline implementation

zonuexe:
[hoaproject/Console: The Hoa\Console library.](https://github.com/hoaproject/Console)

Pure PHP readline implementation.


zonuexe:
I made [zonuexe/psysh-hoa-console](https://github.com/BaguettePHP/psysh-hoa-console) package.

I hope to shrink the package by this pull-request is merged.

bobthecow:
Thank you for this!

I'm a bit reluctant to add `hoa/console` to suggested dependencies, at least with the current text, because most(?) people will not need it. Should it be a runtime warning instead? Maybe more descriptive text would help?

> A pure PHP readline implementation. You'll want this if your PHP install doesn't already support readline or libedit.

But a suggested dependency doesn't help people who use the phar-based install :(

For the phar-based install, I could see building Hoa\Console into the phar by default. It looks like it would increase the size of the file by about 25-30%. I just pushed a commit that offsets it a bit by making the existing phar lighter by 10%, but that's still a non-trivial amount of code to add to support a small subset of users.

Alternatively, we could do what we do when the user is missing `ext-pdo-sqlite`: add it to composer suggested dependencies, but also add a runtime warning for users who run into it with a phar-based install.

#297 has a similar issue, but with `ext-iconv`.

Probably the best thing to do is what was suggested there: to add a more descriptive suggested dependency message to keep users from unintentionally installing this, and also to ship a second (bloated) phar. That phar would include Hoa\Console as a readline polyfill, Symfony's iconv polyfill, and anything else we come across like this.

How's that sound?

src/Psy/Readline/HoaConsole.php
```diff
@@ -0,0 +1,122 @@
...
+use Hoa\Console\Readline\Readline as HoaReadline;
+use Psy\Exception\BreakException;
+
+/**
+ * Hoa\Console Readline implementation.
+ */
+class HoaConsole implements Readline
+{
+    /** @var HoaReadline */
+    private $hoa_readline;
```
bobthecow: Let's call this `$hoaReadline` for consistency with the rest of the codebase.

src/Psy/Readline/HoaConsole.php
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public static function isSupported()
+    {
+        return class_exists('\Hoa\Console\Console', true);
+    }
+
+    public function __construct()
+    {
+        $this->hoa_readline = new HoaReadline();
+        $this->reflection = new \ReflectionProperty($this->hoa_readline, '_history');
```
bobthecow: Can we poke Hoa to get a proper API for history access? I'd rather not depend on reflection on a private member :(

src/Psy/Readline/HoaConsole.php
```diff
@@ -0,0 +1,122 @@
...
+     * Get a STDIN file handle.
+     *
+     * @throws BreakException if user hits Ctrl+D
+     *
+     * @return resource
+     */
+    private function getStdin()
+    {
+        return Console::getInput()->getStream()->getStream();
+    }
```
bobthecow: This looks unused?

src/Psy/Readline/HoaConsole.php
Hywan: What would you like?

Hywan:
Hello 😃,
I am glad to see this PR! Several people asked for a `Hoa\Console` integration inside PsySH. Goals are numerous. `Hoa\Console` is not only a “pure readline implementation” as the description given in this PR suggests, but it is based on terminfo/tput, which makes it interoperable with most of the platforms out there. And regarding readline features, the autocomplete API is powerful and some default autocompleters are already present, like a `Word` or `Path`.
demo

I know some people have developed a shell with `Hoa\Console\Readline` (like https://github.com/jubianchi/atoum-shell cc @jubianchi) if they can help!

composer.json
```diff
@@ -24,13 +24,15 @@
         "phpunit/phpunit": "~3.7|~4.0|~5.0",
         "symfony/finder": "~2.1|~3.0",
         "squizlabs/php_codesniffer": "~2.0",
-        "fabpot/php-cs-fixer": "~1.5"
+        "fabpot/php-cs-fixer": "~1.5",
+        "hoa/console": "~3.16|~1.14"
```
Hywan: Why `~1.14`?

zonuexe: This dependency is added for to support PHP 5.4 and 5.4.
https://travis-ci.org/bobthecow/psysh/builds/122592026

src/Psy/Readline/HoaConsole.php
bobthecow: Looks like we didn't need it. Thanks @zonuexe!

zonuexe:
@bobthecow @Hywan 
Thank you for your review!

bobthecow:
Thanks @zonuexe!

bobthecow:
@Hywan any sense for how much work it'd be to get Hoa\Console autocomplete working with PsySH's current autocompletion? That'd be nice to support as well :)

jubianchi:
this is really great! feel free to ping me for any help.

Hywan:
@bobthecow I have no idea. I don't know how PsySH works. Maybe @zonuexe can give a try. What actually works so far? Is this PR provide readline support with `Hoa\Console` already?

src/Psy/Readline/HoaConsole.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        return $this->hoaReadline->readLine($prompt);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function redisplay()
+    {
+        // noop
```
Hywan: @zonuexe, @bobthecow Actually this is possible to do that. What is the goal of this method exactly?

src/Psy/Readline/HoaConsole.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function listHistory()
+    {
+        $i = 0;
+        $list = array();
+        while (($item = $this->hoaReadline->getHistory($i++)) !== null) {
```
Hywan: @zonuexe @bobthecow We can add a `getWholeHistory`-like method if needed. Thoughts?

bobthecow: It'd be cleaner than this :)

Unfortunately, we need to support all the way back to PHP 5.3, so unless it's ported to the legacy version as well, it won't do us a ton of good.

src/Psy/Readline/HoaConsole.php
bobthecow: `Psy\Readline` is an abstraction over the whole readline API. But `redisplay` isn't _currently_ used in PsySH, so it's more there for completeness than anything else. If Hoa\Console supports it and it's trivial, we should add it for if/whenever PsySH needs to use that.

bobthecow:
@Hywan PsySH currently has readline-based completion support, provided by [AutoCompleter](https://github.com/bobthecow/psysh/blob/master/src/Psy/TabCompletion/AutoCompleter.php). It's not using the Psy\Readline abstraction, but could be changed to use that. If that were the case, it _should_ be as simple as implementing `readline_completion_function` and `readline_info` as part of the Psy\Readline interface.

src/Psy/Readline/HoaConsole.php
Hywan: We can still clear the line and print it again. cc @zonuexe

Hywan:
@bobthecow It's left to @zonuexe to decide whether she/he would like to implement autocompleter too. They are based on the readline though.

src/Psy/Readline/HoaConsole.php
bobthecow: I wouldn't worry about it, personally. it's not really a big deal, since nothing uses it :)

PR #304, opened 2016-05-09
clear code

ReenExe:


bobthecow:
Thanks for your contribution.

Why should we do this? As @GrahamCampbell said in the other PR:

> The reverse refactor would also make a valid PR if you see what I mean.

You could send the opposite commit, with the exact same commit message and pull request title, and it would be just as valid.

Issue #230, opened 2015-08-23
Add automatic update checks


bobthecow:
It's not always obvious when a new version is available (see comments on #225).

Since this package is often installed as a binary, or installed globally, or installed in many different projects, it would be super useful to have some sort of automatic update check. That said, automatic updating is problematic, so let's just add a check and prompt the user to update.

As I see it, there are a few options:
1. Automatically check for updates every time PsySH launches.
2. Automatically check for updates once per day/week/month.
3. Prompt the user to let PsySH check for updates always/daily/weekly/monthly.
4. Don't check for updates ever.

I don't think any of these are mutually exclusive. If there were a couple of configuration options, say:

``` php
[
  'updateCheck' => 'automatic', // or 'prompt', 'never'
  'updateCheckInterval' => 'daily', // or 'always', 'weekly', 'monthly'
]
```

… then we could cover all these options and let the user decide what's best for them.

A sane default would probably be "prompt weekly", but I don't have a strong preference. In the prompt, we could also let the user know about the configuration options:

> It has been over a week since PsySH was updated. Check for updates?
> (To automatically check for updates, or to change how often you're prompted, see …)

Alternatively, on the first prompt we could ask the user what their preferences are and store those.

I'd also probably suppress the check/prompt/etc if PsySH is invoked via `eval(Psy\sh())` (i.e. as an interactive debugger rather than a REPL).

/cc @jimdelois @theofidry

theofidry:
:+1: :)

I would let the default at monthly check: weekly seems a bit too much and it would also be the same frequency as composer check.

GrahamCampbell:
> I would let the default at monthly check

Monthly is way too long IMO. Daily would make more sense.

bobthecow:
If it's automatic, daily is fine. If it's prompting, daily would be super annoying :)

GrahamCampbell:
> If it's prompting, daily would be super annoying :)

Mayby we need an option for people to reply to the prompt with "remind me later"?

theofidry:
Maybe you can do an automatic check daily but in that case it would be better to leave things at showing a notice to tell the user an update is available.

GrahamCampbell:
Yeh, maybe. We'd need to make sure subsequent warnings are unobtrusive though. I'd probably want the initial warning to be quite clear though.

bobthecow:
Once we've established that their current version is outdated, we can just add "Update available" to the header every time it starts. Something like:

```
[...]

PR #314, opened 2016-06-30
Ticket 230

eeree:
Closes bobthecow/psysh#230

I've introduced the checking if there's a new version of PsySH available. It uses tags (as Packagist does) to check if version used by user is up-to-date. 

If executed outside of _cli_ it won't be fired, to prevent checks every time user runs the script on a website.

### DIFF
--- a/src/Psy/Shell.php
+++ b/src/Psy/Shell.php
@@ -17,6 +17,7 @@ use Psy\Exception\Exception as PsyException;
 use Psy\Exception\ThrowUpException;
 use Psy\Output\ShellOutput;
 use Psy\TabCompletion\Matcher;
+use Psy\Util\Cli;
 use Psy\VarDumper\PresenterAware;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command as BaseCommand;
@@ -302,6 +303,7 @@ class Shell extends Application
         // }
 
         $this->output->writeln($this->getHeader());
+        $this->writeVersionInfo();
 
         try {
             $this->loop->run($this);
@@ -882,6 +884,28 @@ class Shell extends Application
                 $this->completion->addMatcher($matcher);
             }
             $this->completion->activate();
+        }
+    }
+
+    /**
+     * @todo Implement self-update
+     * @todo Implement prompt to start update
+     *
+     * @return void|string
+     */
+    protected function writeVersionInfo()
+    {
+        if (!Cli::isCli()) {
+            return;
+        }
+
+        try {
+            $client = $this->config->getChecker();
+            if (!$client->isLatest()) {
+                $this->output->writeln(sprintf('New version is available (current: %s, latest: %s)',self::VERSION, $client->getLatest()));
+            }
+        } catch (\InvalidArgumentException $e) {
+            $this->output->writeln($e->getMessage());
         }
     }
 }
--- a/src/Psy/VersionUpdater/Checker.php
+++ b/src/Psy/VersionUpdater/Checker.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of Psy Shell.
+ *
+ * (c) 2012-2015 Justin Hileman
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy\VersionUpdater;
+
+use Psy\Shell;
+
+class Checker
+{
+    const URL = 'https://api.github.com/repos/bobthecow/psysh/releases/latest';
+
+    private $latest;
+
+    /**
+     * @return bool
+     */
+    public function isLatest()
+    {
+        return version_compare(Shell::VERSION, $this->getVersionFromTag(), '>=');
+    }
+
+    /**
+     * @return string
+     */
+    public function getLatest()
+    {
+        return $this->latest;
+    }
+
+    /**
+     * @param string $version
+     */
+    public function setLatest($version)
+    {
+        $this->latest = $version;
+    }
+
+    /**
+     * @return string|null
+     */
+    private function getVersionFromTag()
+    {
+        $contents = $this->downloadFromGitHub();
+        if (!$contents || !isset($contents->tag_name)) {
+            throw new \InvalidArgumentException('Unable to check for updates');
+        }
+        $this->setLatest($contents->tag_name);
+
+        return $this->getLatest();
+    }
+
+    /**
+     * Set to public to make testing easier.
+     *
+     * @return mixed
+     */
+    public function downloadFromGitHub()
+    {
+        $context = stream_context_create(array('http' => array('user_agent' => 'PsySH/' . Shell::VERSION)));
+
+        return json_decode(@file_get_contents(self::URL, false, $context));
+    }
+}
--- a/test/Psy/Test/ConfigurationTest.php
+++ b/test/Psy/Test/ConfigurationTest.php
@@ -15,6 +15,7 @@ use Psy\CodeCleaner;
 use Psy\Configuration;
 use Psy\ExecutionLoop\Loop;
 use Psy\Output\PassthruPager;
+use Psy\VersionUpdater\Checker;
 use Symfony\Component\Console\Output\ConsoleOutput;
 
 class ConfigurationTest extends \PHPUnit_Framework_TestCase
@@ -237,5 +238,15 @@ class ConfigurationTest extends \PHPUnit_Framework_TestCase
             'invalid color mode: some invalid mode'
         );
         $config->setColorMode($colorMode);
+    }
+
+    public function testSetCheckerValid()
+    {
+        $config = new Configuration();
+        $checker = new Checker();
+
+        $config->setChecker($checker);
+
+        $this->assertSame($checker, $config->getChecker());
     }
 }
--- a/test/Psy/Test/VersionUpdater/CheckerTest.php
+++ b/test/Psy/Test/VersionUpdater/CheckerTest.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
+ * This file is part of Psy Shell.
+ *
+ * (c) 2012-2015 Justin Hileman
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy\Test\VersionUpdater;
+
+use Psy\Shell;
+
+class CheckerTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @dataProvider malformedResults
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessage Unable to check for updates
+     *
+     * @param $input
+     */
+    public function testExceptionInvocation($input)
+    {
+        $checker = $this->getMockBuilder('Psy\\VersionUpdater\\Checker')
+            ->setMethods(array('downloadFromGitHub'))
+            ->getMock();
+        $checker->expects($this->once())->method('downloadFromGitHub')->willReturn($input);
+        $checker->isLatest();
+    }
+
+    /**
+     * @dataProvider jsonResults
+     *
+     * @param $assertion
+     * @param $input
+     */
+    public function testDataSetResults($assertion, $input)
+    {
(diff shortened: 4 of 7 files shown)


eeree:
It's just an initial commit. What have to be done:
* configuration options have to be applied
* some temporary storage for last check time
* self-update
* _--version update_ method
* move from factory to DiC (needs to be applied in a whole project)

src/Psy/Shell.php
```diff
@@ -302,6 +305,7 @@ public function doRun(InputInterface $input, OutputInterface $output)
         // }
 
         $this->output->writeln($this->getHeader());
+        $this->output->writeln($this->getVersionCheck());
```
bobthecow: this probably doesn't need to write the output, since you already did it inside `getVersionCheck`?

bobthecow:
Thanks for getting this started.

src/Psy/Shell.php
bobthecow: this probably doesn't need to write the output, since you already did it inside `getVersionCheck`?

src/Psy/VersionUpdater/CheckerFactory.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy\VersionUpdater;
+
+use GuzzleHttp\Client;
+
+class CheckerFactory
```
bobthecow: `Configuration` acts as our factory and dependency injection container, especially for simple things like this.

composer.json
```diff
@@ -13,12 +13,13 @@
         }
     ],
     "require": {
-        "php": ">=5.3.9",
+        "php": ">=5.5.0",
```
bobthecow: i'm afraid this change will be a deal breaker.

screen shot 2016-06-30 at 7 30 13 pm

src/Psy/VersionUpdater/Checker.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function isLatest()
+    {
+        try {
+            $this->current = new Version(Shell::VERSION);
+            $this->latest = new Version($this->getTagName());
+
+            return $this->current->compare($this->latest);
+        } catch (\InvalidArgumentException $e) {
+            return true;
```
bobthecow: if it fails, "unable to check for updates" is probably a better message than "yep! totally up to date!"

bobthecow:
Also, Guzzle is a pretty hefty dependency for effectively doing the same thing as this:

```
$url = 'https://api.github.com/repos/bobthecow/psysh/releases/latest';
$context = stream_context_create(['http' => ['user_agent' => 'PsySH/0.7.2']]);
$contents = json_decode(file_get_contents($url, false, $context));
```

composer.json
Taluu: But php 5.5 is almost dead and won't receive any updates really soon. "Forcing" an update to at ;east 5.6 (the ones stuck at < 5.6 can still use previous versions, that's all) is IMO the best factor to motivate upgrades...

Otherwise, let'sjust support 5.2 as there are still 10% usage :trollface:

Taluu:
I ranted about the php version, but I agree that guzzle isn't really necessary, a simple fgc is more than enough

composer.json
bobthecow: 10% use 5.2.

80% use <= 5.5. That's a _huge_ difference.

Taluu: my point is still valid : following and supporting * stable * versions of php is what will help deprecated / unmaintained versions to stay what they should be, dead.

As you are following semver (and even without that actually, as in the semver, it doesn't say you should bump major versions on dependency upgrades...), if someone wants to have a psysh that supports an old version of php, they should just use a previous version that supports it. Then it's up to you to keep or not a support (if it was up to me, well... there should not be any, except one or two previous versions at most)

Especially as psysh is a dev tool. But once again, it's only my opinion, I'll still be using psysh, even though I may not agree with some points, as it is still awesome :}

Taluu: p.s : these stats are meaningless IMO, as the percentages of versions usage on websites is not quite the same as the percentage of usage of php versions and psysh usage.... especially as A LOT of providers are NOT upgrading and letting sites rot on old php versions.

theofidry: According to [Jordi's stats](https://seld.be/notes/php-versions-stats-2016-1-edition), PHP <= 5.5 is more around 40% than 80 (in May 2016). Plus it's not like bumping the PHP version requirement is gonna break old releases... As @Taluu it's also the responsibility of library maintainers to push libraries. If there's always a good support for older versions, nobody is going to upgrade.

GrahamCampbell: PHP 5.5 is EOL now too. Everyone should be on PHP 5.6+ unless they are making security fixes and building PHP themselves.

GrahamCampbell: I'd recommend bumping to 5.5.9 actually.

composer.json
```diff
@@ -13,12 +13,13 @@
...
         "symfony/console": "~2.3.10|^2.4.2|~3.0",
         "symfony/var-dumper": "~2.7|~3.0",
         "nikic/php-parser": "^1.2.1|~2.0",
         "dnoegel/php-xdg-base-dir": "0.1",
-        "jakub-onderka/php-console-highlighter": "0.3.*"
+        "jakub-onderka/php-console-highlighter": "0.3.*",
+        "guzzlehttp/guzzle": "^6.0"
```
GrahamCampbell: You should allow the use of guzzle 5 too to avoid conflicting with everything.

GrahamCampbell: That would allow us to only have to bump to PHP 5.4 too.

test/Psy/Test/ConfigurationTest.php
```diff
@@ -238,4 +239,22 @@ public function testSetColorModeInvalid()
         );
         $config->setColorMode($colorMode);
     }
+
+    public function testSetCheckerInvalid()
+    {
+        $config = new Configuration();
+
+        $this->setExpectedException('TypeError');
```
bobthecow: The rest of the project uses the annotation style for expected exceptions. I don't have a strong preference, but we should prolly keep it consistent :)

test/Psy/Test/VersionUpdater/CheckerTest.php
```diff
@@ -0,0 +1,81 @@
...
+    public function malformedResults()
+    {
+        return [
+            [null],
+            [false],
+            [true],
+            [json_decode('{"foo":"bar"}')],
+            [json_decode('{}')],
+            [json_decode('[]')],
+            [   []],
```
bobthecow: something whitespacey happened here.

src/Psy/Shell.php
```diff
@@ -11,13 +11,16 @@
...
+use DoctrineTest\InstantiatorTestAsset\ExceptionAsset;
 use Psy\Exception\BreakException;
 use Psy\Exception\ErrorException;
 use Psy\Exception\Exception as PsyException;
 use Psy\Exception\ThrowUpException;
 use Psy\Output\ShellOutput;
 use Psy\TabCompletion\Matcher;
+use Psy\Util\Cli;
 use Psy\VarDumper\PresenterAware;
+use Psy\VersionUpdater\CheckerFactory;
```
bobthecow: leftover `use` statement.

src/Psy/Shell.php
```diff
@@ -11,13 +11,16 @@
 
 namespace Psy;
 
+use DoctrineTest\InstantiatorTestAsset\ExceptionAsset;
```
bobthecow: leftover `use` or… ?

test/Psy/Test/ConfigurationTest.php
bobthecow: Oh, this is checking for a TypeError. This isn't going to work on anything less than PHP 7.

test/Psy/Test/VersionUpdater/CheckerTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->assertSame($assertion, $checker->isLatest());
+    }
+
+    /**
+     * @return array
+     */
+    public function jsonResults()
+    {
+        return [
+            [false, json_decode('{"tag_name":"v9.0.0"}')],
```
bobthecow: shorthand array syntax doesn't work in PHP 5.3

side note: if someone wants to make an argument for dropping support for older PHP, this would be way more compelling to me than "it's deprecated so we should force users to upgrade" :P

bobthecow:
Just to save you from some more code style issues, I'm going to merge #316. Do you mind squashing and rebasing after that so we can get a clean diff to work with?

eeree:
Sure thing. I will do the rebase tomorrow then, as it's almost 1 am here.

bobthecow:
No worries. It's merged now :)

Have a good night!

test/Psy/Test/VersionUpdater/CheckerTest.php
theofidry: @bobthecow drop support for PHP < 5.6 :P

bobthecow: Not going to happen.

Taluu: Here goes...

- PHP 5.5 is now officially dead for some weeks... php 5.3 is not receiving support for something like 5 (!!) years.
- Still the ramblings about the fact that's it up to lib développer to upgrade and thus promote new versions. Why bother when everything works nicely with the same setup ?
- no need to do itsy beetsy things to go around deprecated or now inexistant limitations of older versions.
-Once again, psysh is a **dev** tool... nowadays, you really have to be fucked up for not having at least 5.5 in your local machine.
- Awesome features in recent versions, such as traits, optimizations (see above), generators, awesome autoloading in composer since 5.6, constant resolved at execution, ... and so on

Sorry for the langage on some points (especially before the last one, in which i don't want to be offensive (Well okay maybe a little :D )), but there you go. :}

src/Psy/Shell.php
```diff
@@ -302,6 +303,7 @@ public function doRun(InputInterface $input, OutputInterface $output)
         // }
 
         $this->output->writeln($this->getHeader());
+        $this->output->writeln($this->getVersionCheckResult());
```
bobthecow: Does this add an extra newline when there's nothing to report from the version checker?

eeree: @bobthecow Yes it does. Do you think it shouldn't happen?

bobthecow: I'd prefer if it didn't :)

bobthecow:
Hmm. Looking at `Console\Output`, it looks like `writeLn(null)` and `writeLn([])` don't do anything anyway, so it shouldn't have written an empty line if `getVersionCheckResult` returned null. I wouldn't have guessed that from the function name :)

eeree:
@bobthecow Damn, you're right. I didn't think about it at the moment of writing the change as I've been focused on empty line. What's your advice?

eeree:
Hmmm... what about introducing the check by interval? Check once per week by default (configurable in _config.php_)? I will just have to check how does _psysh_ cache the results and it can be done in another PR?

bobthecow:
Yep, it's fine to do that as a separate PR. Want to squash and rebase this against develop and I'll merge?

For the interval, a 1 week default seems reasonable. Maybe support "always", "daily", "weekly" and "never"?

For caching, I'd just go with a JSON file in `$XDG_CACHE_HOME` / `.psysh` / `$APPDATA/PsySH`. See ConfigPaths for how we deal with other config directory types.

eeree:
Sure thing! Rebased and squashed.

I'll do the _config.php_ part during the week.

bobthecow:
Thanks, merged!

GrahamCampbell:
Is there a way to disable this (say for if bundled with laravel)?

bobthecow:
See the last few comments.  This isn't going to land in a stable release without the ability to disable it.

eeree:
Correct. It's merged for _testing_ purposes only. Next task is to introduce frequency of scans. Probably in case of cache unavailability scans also shouldn't happen (to prevent slow downs).

PR #325, opened 2016-10-22
Support php-parser v3

GrahamCampbell:


bobthecow:
Does PHP 7.1 not have an xdebug config to disable? can we make that line more robust, so it disables if it detects the config, rather than special casing versions we don't think will have it?

bobthecow:
```
(php --ini | grep xdebug) && phpenv config-rm xdebug.ini
```

?

GrahamCampbell:
NB Currently this code never actually resolves php parser 3.x since it's not stable.

bobthecow:
oh, good to know. is 3.x required for 7.1?

GrahamCampbell:
> oh, good to know. is 3.x required for 7.1?

Yeh, php parser 2.x doesn't parse php 7.1 code, though can parse older code on php 7.1 just fine.

bobthecow:
🆒

bobthecow:
oh, that xdebug thing prolly needs an `|| true` on the end to keep it from barfing when the result of grep is false.

GrahamCampbell:
Ok, rebased this on develop.

GrahamCampbell:
Ok, PHP-Parser made some breaking changes. Will investigate.

GrahamCampbell:
@bobthecow Looks good now. This is needed in order to get Laravel properly onto PHP 7.1 since class preloader also depends on PHPParser.

bobthecow:
Thanks! Any ideas why hhvm didn't finish? The xdebug change?

bobthecow:
```
[[ "$TRAVIS_PHP_VERSION" != hhvm* ]] && (php --ini | grep xdebug) && phpenv config-rm xdebug.ini || true
```

… ought to do the trick?

GrahamCampbell:
Passing now. :)

bobthecow:
Awesome. Thank you.

Issue #67, opened 2014-04-23
PHP is not resumed when using built-in web server


ghost:
When using the built-in web server of PHP, psysh is executed when using `\Psy\Shell::debug(get_defined_vars());`, however the PHP execution is not resumed after quitting psysh (the browser hang waiting for an answer from the server). When launching a PHP script (with the web server), the PHP script is resumed after hitting ctrl-d in psysh.

I'm using the master branch of psysh on Ubuntu 14.04 with symfony2.

```
php --version

PHP 5.5.9-1ubuntu4 (cli) (built: Apr  9 2014 17:11:57) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
```

moolen:
+1

Im actually getting dropped into a second psysh shell but it takes about 60 seconds.

im running `PHP 5.5.9-1ubuntu4.4 (cli) (built: Sep  4 2014 06:56:34)`

The script to reproduce:
```PHP
<?php

require 'vendor/autoload.php';

$first = "foo";
\Psy\Shell::debug(get_defined_vars());

$second = "bar";
\Psy\Shell::debug(get_defined_vars());

echo $first . "\n";
echo $second . "\n";
```

It takes about 60s to get to the second shell.
About 60 seconds after i exited the second shell the output is transferred to the client, but the connection never gets closed.

moolen:
a workaround: set the `default_socket_timeout` in your `/etc/php5/cli/php.ini` to 1 second.
so you have a lower latency between the breakpoints...

bobthecow:
Oh, interesting. Thanks for that @moolen!

oniric85:
+1 I've the same problem, is there any update on this? I can help if you need more information.

ZalmanBlau:
+1

ricpelo:
+1 here too.

bobthecow:
Thanks to @5tevan, this is fixed in the development branch. It'll go out in the next release :)

andreiglingeanu:
Amazing!

bobthecow:
This should be fixed (as of v0.8.0). Let me know if you run into any issues.

PR #326, opened 2016-10-28
Prevent PHP's built-in web server from hanging

5tevan:
This should fix the issue in #67. `pcntl_fork` doesn't work well with PHP's built-in web server. I believe this has to do with the `exit` call in the child cleaning up resources its parent still needs.

**Caveat**: When running this patch under PHP 5.6.26 with `tabCompletion => true` a segfault occurs on the second request to the server during the call to `readline_completion_function` in `Psy\TabCompletion\AutoCompleter::activate`. This issue appears fixed in PHP 7.

### DIFF
--- a/src/Psy/ExecutionLoop/ForkingLoop.php
+++ b/src/Psy/ExecutionLoop/ForkingLoop.php
@@ -82,7 +82,7 @@ class ForkingLoop extends Loop
         fwrite($up, $this->serializeReturn($shell->getScopeVariables()));
         fclose($up);
 
-        exit;
+        posix_kill(posix_getpid(), SIGKILL);
     }
 
     /**


GrahamCampbell:
Why would you be running this on a web server?

bobthecow:
@GrahamCampbell Because if you spin up the built-in webserver from the command line, you can put in an `eval(\Psy\sh())` call in a request handler, interrupt the response, and mess with things in realtime. It's pretty ok :)

GrahamCampbell:
Interesting. I was just curious as to why you'd do that. :)

bobthecow:
Awesome. Works like a charm. Thanks @5tevan!

PR #331, opened 2016-11-30
Add Yii Framework 2 (yii2-shell) to "As Seen On"

thyseus:


bobthecow:
This is already fixed in the dev branch. I just need to cut a new release so it shows up in master :)

PR #339, opened 2016-12-20
Fix “No such file or directory” update_check.json

zonuexe:
```
% php ./my-shell.php
Psy Shell v0.8.0 (PHP 5.5.26 — cli) by Justin Hileman
ErrorException with message 'file_get_contents(/home/tadsan/.config/psysh/update_check.json): failed to open stream: No such file or directory'
Bye...
```

### DIFF
--- a/src/Psy/ConfigPaths.php
+++ b/src/Psy/ConfigPaths.php
@@ -180,4 +180,41 @@ class ConfigPaths
 
         return $files;
     }
+
+    /**
+     * Touch $file mkdir and if $dir is not exists.
+     *
+     * Generates E_USER_NOTICE error if $file or $dir is not writable.
+     *
+     * @param string $dir
+     * @param string $file
+     *
+     * @return string|false Full path to $file. Return false if file is not writable.
+     */
+    public static function touchFileWithMkdir($dir, $file)
+    {
+        if (!is_writable($dir)) {
+            trigger_error("Writing to {$dir} is not allowed.", E_USER_NOTICE);
+
+            return false;
+        }
+
+        if (!is_dir($dir)) {
+            mkdir($dir, 0700, true);
+        }
+
+        $path = sprintf('%s/%s', rtrim($dir, '/'), ltrim($file, '/'));
+
+        if (!is_writable($path)) {
+            trigger_error("Writing to {$path} is not allowed.", E_USER_NOTICE);
+
+            return false;
+        }
+
+        if (!file_exists($path)) {
+            touch($path);
+        }
+
+        return $path;
+    }
 }
--- a/src/Psy/ConfigPaths.php
+++ b/src/Psy/ConfigPaths.php
@@ -193,6 +193,18 @@ class ConfigPaths
      */
     public static function touchFileWithMkdir($dir, $file)
     {
+        $path = sprintf('%s/%s', rtrim($dir, '/'), ltrim($file, '/'));
+
+        if (file_exists($path)) {
+            if (is_writable($path)) {
+                return $path;
+            }
+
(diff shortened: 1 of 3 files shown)


src/Psy/Configuration.php
```diff
@@ -1110,7 +1110,12 @@ public function getChecker()
                 case Checker::DAILY:
                 case Checker::WEEKLY:
                 case Checker::MONTHLY:
-                    $this->checker = new IntervalChecker($this->getUpdateCheckCacheFile(), $interval);
+                    $check_file = $this->getUpdateCheckCacheFile();
+                    if ($check_file === false) {
+                        $this->checker = new NoopChecker();
```
bobthecow: i think falling back to a `NoopChecker` when the user hasn't explicitly turned off update checking is probably grounds for showing an error message or warning of some sort.

src/Psy/Configuration.php
```diff
@@ -1162,15 +1167,30 @@ public function setUpdateCheck($interval)
...
      * Get a cache file path for the update checker.
      *
-     * @return string
+     * @return string|false Return false if config file/directory is not writable
      */
     public function getUpdateCheckCacheFile()
     {
         $dir = $this->configDir ?: ConfigPaths::getCurrentConfigDir();
+
+        if (!is_writable($dir)) {
```
bobthecow: this problem needs to be solved for the history file as well. any thoughts on a more general solution?

src/Psy/ConfigPaths.php
```diff
@@ -180,4 +180,41 @@ private static function getRealFiles(array $dirNames, array $fileNames)
...
+     *
+     * @param string $dir
+     * @param string $file
+     *
+     * @return string|false Full path to $file. Return false if file is not writable.
+     */
+    public static function touchFileWithMkdir($dir, $file)
+    {
+        if (!is_writable($dir)) {
+            trigger_error("Writing to {$dir} is not allowed.", E_USER_NOTICE);
```
bobthecow: If `$path` exists and is writable, but `$dir` isn't writable, we probably shouldn't treat that as an error.

src/Psy/Configuration.php
```diff
@@ -1110,7 +1112,12 @@ public function getChecker()
                 case Checker::DAILY:
                 case Checker::WEEKLY:
                 case Checker::MONTHLY:
-                    $this->checker = new IntervalChecker($this->getUpdateCheckCacheFile(), $interval);
+                    $check_file = $this->getUpdateCheckCacheFile();
```
bobthecow: Please use `$checkFile` for consistency.

src/Psy/Configuration.php
```diff
@@ -1162,15 +1169,12 @@ public function setUpdateCheck($interval)
...
      */
     public function getUpdateCheckCacheFile()
     {
         $dir = $this->configDir ?: ConfigPaths::getCurrentConfigDir();
-        if (!is_dir($dir)) {
-            mkdir($dir, 0700, true);
-        }
 
-        return $dir . '/update_check.json';
+        return ConfigPaths::touchFileWithMkdir($dir, '/update_check.json');
```
bobthecow: Leave the slash off the file name?

zonuexe: Fixed.

src/Psy/Configuration.php
zonuexe: Fixed.

bobthecow:
The `is_writable` check was in the wrong place, because a directory that doesn't exist by definition isn't writable, which is why CI was failing. But I've merged this PR along with a fix. Thanks!

zonuexe:
🙇

PR #354, opened 2017-02-08
Allow running in the global scope

### LINKED ISSUE #353: Support for global scope
It would be great if there was a way to set the shell to use the global scope (for debugging legacy applications which heavily rely on global configuration). `Shell::setScopeVariables()` can load the values from the global scope, but does not actually give access to it.

Downstream patch where this came up: https://gerrit.wikimedia.org/r/#/c/334217/

tgr:
Add Shell/Context option to use the global scope. When set, ignore scope variables and copy (by reference) `$GLOBALS` into the scope.
    
Fixes #353.

This is moderately horrible; I don't think there is a non-horrible way of doing it though.

### DIFF
--- a/src/Psy/Context.php
+++ b/src/Psy/Context.php
@@ -21,6 +21,7 @@ class Context
 {
     private static $specialVars = array('_', '_e', '__psysh__', 'this');
     private $scopeVariables = array();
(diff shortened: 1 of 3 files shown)


bobthecow:
Closing in favor of the response in #353

PR #358, opened 2017-02-22
Catch PHP7 \Errors in ForkingLoop::serializeReturn()

damiankloip:
In Drush we have this problem https://github.com/drush-ops/drush/pull/2635 . If/when the Drupal container gets serialized and assertions are enabled, we get an `AssertionError` but these are not caught as they extend `Error` (and implement `Throwable`).

Does it makes sense to add this here instead? When testing it seems to make no difference when using PHP5.6 as it should always get the `Exception` first anyway.

### DIFF
--- a/src/Psy/ExecutionLoop/ForkingLoop.php
+++ b/src/Psy/ExecutionLoop/ForkingLoop.php
@@ -166,8 +166,7 @@ class ForkingLoop extends Loop
                 $serializable[$key] = $value;
             } catch (\Exception $e) {
                 // we'll just ignore this one...
-            }
-            catch (\Error $e) {
+            } catch (\Error $e) {
                 // and this one too...
             }
         }
--- a/src/Psy/ExecutionLoop/ForkingLoop.php
+++ b/src/Psy/ExecutionLoop/ForkingLoop.php
@@ -166,7 +166,7 @@ class ForkingLoop extends Loop
                 $serializable[$key] = $value;
             } catch (\Exception $e) {
                 // we'll just ignore this one...
-            } catch (\Error $e) {
+            } catch (\Throwable $e) {
                 // and this one too...
             }
         }
--- a/src/Psy/ExecutionLoop/ForkingLoop.php
+++ b/src/Psy/ExecutionLoop/ForkingLoop.php
@@ -167,6 +167,9 @@ class ForkingLoop extends Loop
             } catch (\Exception $e) {
                 // we'll just ignore this one...
             }
+            catch (\Error $e) {
+                // and this one too...
+            }
         }
 
         return @serialize($serializable);


bobthecow:
It makes perfect sense to add here. PHP's really cool about catching things it knows nothing about. We already added handling for `\Error` (and `\TypeError`) inside the loop:
 
https://github.com/bobthecow/psysh/blob/701e8a1cc426ee170f1296f5d9f6b8a26ad25c4a/src/Psy/ExecutionLoop/Loop.php#L108

… but I guess I'd never serialized anything that threw an `\Error` :)

bobthecow:
Do you mind switching the PR target to the `develop` branch?

src/Psy/ExecutionLoop/ForkingLoop.php
```diff
@@ -166,6 +166,8 @@ private function serializeReturn(array $return)
                 $serializable[$key] = $value;
             } catch (\Exception $e) {
                 // we'll just ignore this one...
+            } catch (\Error $e) {
```
Taluu: instead you should catch `Throwable $e`

bobthecow: Good catch! You've still got to catch `\Exception` for PHP 5.x, though, so it doesn't make a huge difference.

damiankloip: Yeah, plus in the regular `Loop` class it catches `Error` too. @bobthecow Do you want me to change it to throwable?

bobthecow: The regular loop does different things with `Error` vs `TypeError` vs `Exception`, so let's leave it there.

damiankloip: No, I mean this `ForkingLoop` hunk :)

bobthecow:
Oh, right. Those merge commits. Grr.

damiankloip:
Don't worry, rebasing onto develop now, we should be getting rid of those pesky merge commits any second.

Taluu:
interactive rebase o

(imo you should drop the dev / master duality branches, and use only the master for dev, especially as you already have the release/x branches...)

src/Psy/ExecutionLoop/ForkingLoop.php
bobthecow: Sure. Might as well use `Throwable` here. It should be more future proof. You know, in case PHP 8 decides we should be able to throw more than just Errors and Exceptions :)

damiankloip:
@Taluu I can just use `git rebase --onto develop master` and be done with it. Interactive rebase I'd have to do more work for sure.

Taluu:
yup, just removing the commits, and that's pretty much it. But we digress, and `--onto` is also a good solution :'}

bobthecow:
Thanks! This will go into the next stable release, which should be in the next couple of days.

damiankloip:
Sounds great. Thanks @bobthecow

PR #359, opened 2017-02-22
Better VarDumper escaping.

### LINKED ISSUE #352: PsySH stuck on specific strings.
One day I've been peeking to a table containing binary data. I've expected PsySH to print me a pile of random characters, but instead it became stuck throwing me "Incorrectly nested style tag found." on every next input line.

The fastest way to reproduce it is trying to dump ```"\\\n"```

```
Psy Shell v0.8.1 (PHP 7.1.0 — cli) by Justin Hilemanan
>>> 2*3
=> 6
>>> "\\\n"
Symfony\Component\Console\Exception\InvalidArgumentException with message 'Incorrectly nested style tag found.'
>>> 2*3
Symfony\Component\Console\Exception\InvalidArgumentException with message 'Incorrectly nested style tag found.'
```

bobthecow:
In the current implementation, strings with a literal backslash which are immediately followed by a control character get formatted as:

    \<cchr>\n</cchr>

… which actually escapes the opening tag for cchr. And this screws up Symfony's `OutputFormatter`. Forever, apparently.

cc @nicolas-grekas

Fixes #352

### DIFF
--- a/src/Psy/VarDumper/Dumper.php
+++ b/src/Psy/VarDumper/Dumper.php
@@ -22,7 +22,8 @@ class Dumper extends CliDumper
 {
     private $formatter;
 
-    protected static $controlCharsRx = '/[\x00-\x1F\x7F]+/';
+    protected static $onlyControlCharsRx = '/^[\x00-\x1F\x7F]+$/';
+    protected static $controlCharsRx = '/([\x00-\x1F\x7F]+)/';
     protected static $controlCharsMap = array(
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Good catch. Symfony's CliDumper is free from this issue because it deals with its own color theme and doesn't use the console formatter...

bobthecow:
Yeah. I've been considering going that route as well :-/

PR #374, opened 2017-05-20
Remove PHP 5.3 support

castarco:
The main intent is simplifying the code, but also to remove support for PHP < 5.4 .

Rationale:
  * Improve maintainability (because we have less code and less conditions)
  * Improve security (PHP 5.3 is insecure because a lot of reasons)
    * https://secure.php.net/supported-versions.php 
  * PHP <= 5.3 is no longer used, at least by Composer users:
    * https://seld.be/notes/php-versions-stats-2017-1-edition
  * Open the door to use features that are "only" available on PHP >= 5.4 .
  * Improve CI times.

This PR makes the assumption that every sensible user of PsySH will be using relatively new versions, but without going "too far" (in fact, this PR is very conservative), so PHP 5.4 is the minimum expected version.


src/Psy/CodeCleaner/LoopContextPass.php
```diff
@@ -28,12 +28,10 @@
  */
 class LoopContextPass extends CodeCleanerPass
 {
-    private $isPHP54;
     private $loopDepth;
 
     public function __construct()
```
GrahamCampbell: No need for this constructor now

src/Psy/CodeCleaner/StaticConstructorPass.php
```diff
@@ -29,12 +29,10 @@
  */
 class StaticConstructorPass extends CodeCleanerPass
 {
-    private $isPHP533;
     private $namespace;
 
     public function __construct()
```
GrahamCampbell: same

castarco: done

src/Psy/CodeCleaner/LoopContextPass.php
castarco: done

bobthecow:
:+1:

I'm planning on dropping PHP 5.3 support with the 0.9 release, but haven't wanted to make a backwards compatibility break just for the sake of this, so I haven't yet :)

src/Psy/CodeCleaner/StaticConstructorPass.php
```diff
@@ -55,7 +49,7 @@ public function enterNode(Node $node)
             $this->namespace = isset($node->name) ? $node->name->parts : array();
         } elseif ($node instanceof Class_) {
             // Bail early if this is PHP 5.3.3 and we have a namespaced class
-            if (!empty($this->namespace) && $this->isPHP533) {
```
GrahamCampbell: this change is wrong

GrahamCampbell: ah, wait, no, just confusing variable name :)

GrahamCampbell: the comment still needs editing though

castarco: Done

composer.json
```diff
@@ -13,18 +13,18 @@
         }
     ],
     "require": {
-        "php": ">=5.3.9",
-        "symfony/console": "~2.3.10|^2.4.2|~3.0",
+        "php": ">=5.4.0",
+        "symfony/console": "^2.4.2|~3.0",
         "symfony/var-dumper": "~2.7|~3.0",
-        "nikic/php-parser": "~1.3|~2.0|~3.0",
```
GrahamCampbell: With this change, is there any version dependent php parser code that can go?

Taluu:
How about going to PHP7 ? PHP 5.4 is dead too (and PHP 5.5 too). And according to the composer stats, more than 50% are using at least PHP 7.0 (Reminder : This is s a dev tool, so even the composer stats may not be really on par).

GrahamCampbell:
> How about going to PHP7

👎 Don't see the need for this package to kill support tbh. Maybe in another 2 years. I think php 5.4 is fine in this instance.

GrahamCampbell:
> This is s a dev tool

Not really, actually. I use this in production with Laravel apps all the time.

Taluu:
You're debugging on live prod ? Logging in CLI on prod ? That sounds like some really horrible ideas...

5.4 has been given the death flag for something like 2 years, and php 5.5 for something like a year. At least consider making this 5.6...

Anyway, as it's `0.9`, BC Breaks are tolerated (that's the whole point of semantic versioning). And those who don't use the required php version, they can stick to 0.8 (composer's solver is smart, you know). If a cool new feature got your interest, you can either upgrade your env or backport it (but then the support doesn't fall to this package maintainers)...

GrahamCampbell:
> You're debugging on live prod ? Logging in CLI on prod ? That sounds like some really horrible ideas...

I don't see anything wrong with making database queries on production, through the `artisan tinker` command. Often my apps are architected using a command bus, so it's not uncommon for me to dispatch commands, using the command line, for things that do not exist on the admin web interface yet.

composer.json
castarco: I don't understand the question.

castarco:
@bobthecow I think that removing PHP 5.3 support is not only for the sake of doing it. There are multiple and serious reasons for doing it.

PHP 5.3 support ended more than 3 years ago (including security fixes), almost every important project that depends on PsySH has bumped its minimum requirements.

I think that the same reasoning applies to PHP 5.4 & 5.5, but I prefer to be conservative on this regard, and this is the reason why I only removed PHP 5.3 support in this PR.

bobthecow:
I meant more that I want to batch several backwards compatibility breaks in one. So v0.9.0 removes PHP 5.3 support _and_ cleans up other things that need it.

bobthecow:
Do you mind updating the base to the `0.9.x-dev` branch?

bobthecow:
>At least consider making this 5.6...

Projects should be strict about supporting only the newest PHP and dependency versions.

Frameworks should be opinionated and forward looking, encouraging projects to update to the newest PHP and dependency versions.

Libraries should be as compatible with as many versions of PHP and their dependencies as is reasonable to support. 

PsySH is a library 😃

castarco:
Ok, in a few minutes this PR will be retargeted to the 0.9 branch.

castarco:
Done.

bobthecow:
Thanks!

PR #375, opened 2017-05-25
Loop listeners

bobthecow:
The first of the backwards compatibility breaks I'd like to see in v0.9.x.

This replaces the `Loop`/`ForkingLoop` concept with a variation on the listeners that @nateabele implemented for runkit-based code reloading. With this change, anyone supplying their own custom `Loop` implementation will have to re-implement it in terms of `ExecutionLoop\Listener`. As far as I can tell, this will affect approximately zero people, but someone with better GitHub code search foo can confirm :)

CodeCleaner and command execution might both be good candidates for loop listener conversion as well.

Loop listeners aren't (yet) configurable, but in some conceivable future they could be.

As an added bonus, this change will (finally) ship code reloading… but only to people running PHP 5.x. 🎉 ‼️ ☹️ ‼️

### DIFF
--- a/src/Psy/ExecutionLoop/RunkitReloader.php
+++ b/src/Psy/ExecutionLoop/RunkitReloader.php
@@ -3,13 +3,13 @@
 /*
  * This file is part of Psy Shell.
  *
- * (c) 2012-2015 Justin Hileman
+ * (c) 2012-2017 Justin Hileman
  *
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
 
-namespace Psy\Listener;
+namespace Psy\ExecutionLoop;
 
 use Psy\Exception\ParseErrorException;
 use Psy\ParserFactory;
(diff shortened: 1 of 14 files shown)


bobthecow:
@castarco Take a look at this? It's a backwards compatibility break, but as far as I can tell almost nobody is affected by it. Jupyter-PHP might be the exception.

I looked through the code, and I _think_ the only thing you'll need to change is the class name for the constant [on this line](https://github.com/Litipk/Jupyter-PHP/blob/80b7db1dccf032e282e71fa87e1266291d12db6a/src/Actions/ExecuteAction.php#L115)… Am I missing anything? Is removing the configurable loop going to cause you any trouble?

castarco:
@bobthecow go ahead, I'll adapt the code :+1:

GrahamCampbell:
Looks pretty good. :+1:

src/Psy/ExecutionLoop/RunkitReloader.php
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @return bool
+     */
+    public static function isSupported()
+    {
+        return extension_loaded('runkit');
+    }
+
+    /**
+     * Constructor.
```
bobthecow: :)

bobthecow: Fixed!

PR #376, opened 2017-05-26
Update README.md

GrahamCampbell:


bobthecow:
Thanks! This has been merged into `develop`.

bobthecow:
I'm not super familiar with the Laravel/Lumen relationship, but it looks like `laravel/tinker` isn't a direct replacement, as Lumen itself doesn't depend on `illuminate/console`, and its `artisan` uses `symfony/console` directly instead.

Should we keep links to both packages around, or does anyone who wants `tinker` just have to incur an additional dependency?

Is there any advantage to maintaining a separate and more specialized Lumen-specific `tinker`?

GrahamCampbell:
> I'm not super familiar with the Laravel/Lumen relationship, but it looks like laravel/tinker isn't a direct replacement, as Lumen itself doesn't depend on illuminate/console, and its artisan uses symfony/console directly instead.

It is a replacement, and it ships with the Lumen core. :)

GrahamCampbell:
(Oh, we changed our mind on putting it in lumen core, but I can assure you it is the official package for both Laravel and Lumen - I wrote it myself)

GrahamCampbell:
The whole point of us having a laravel/tinker package was to:

1. Make it available in lumen.
2. Make it an optional dependency in Laravel, since php parser is jumping through versions very quickly, so not everyone can keep up.

GrahamCampbell:
Also, we put it in the offical composer suggest: https://github.com/laravel/lumen-framework/blob/5.4/composer.json#L51

bobthecow:
Great. Thanks for the context :)

GrahamCampbell:
Is this going to merged then, since laravel/tinker is the official package for lumen.

bobthecow:
Yep. It's already merged into develop (see above) and will show up in master with the next release.

GrahamCampbell:
Ah, cool. 👍

PR #377, opened 2017-05-26
Support `xxxxx or die` syntax

zonuexe:
`foo() or die(1);` and `$f = function () { 1 or die; };` are valid statements, but it is not accepted by PsySH.
Because `foo or throw new Exception()` is not a valid statement.

```
PHP error:  syntax error, unexpected throw (T_THROW) on line 2
```

This constraint can be avoided by a trick that combines Closure and `call_user_func`.

### DIFF
--- a/src/Psy/CodeCleaner/ExitPass.php
+++ b/src/Psy/CodeCleaner/ExitPass.php
@@ -12,12 +12,9 @@
 namespace Psy\CodeCleaner;
 
 use PhpParser\Node;
-use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\Exit_;
-use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Name\FullyQualified as FullyQualifiedName;
-use PhpParser\Node\Scalar\String_;
-use PhpParser\Node\Stmt\Throw_;
 
 class ExitPass extends CodeCleanerPass
 {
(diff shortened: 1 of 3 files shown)


bobthecow:
This should be generalized to more cases. `die` (and `exit`) is an expression and can be used anywhere an expression could be used. `throw` is a statement, so it basically only works on its own line. Other perfectly valid PHP cases that fail:

```php
if (exit) { echo "wat"; }
exit or die;
switch (die) { }
for ($i = 1; $i < 10; die) {}
```

I mean, lots of these sort of cases make no sense, but ¯\_(ツ)_/¯

We just need our replacement to always be an expression rather than a statement. So that simplifies this diff by removing any special casing.

If we add a `BreakException::exit` static method, which simply throws a breakException, it's even easier :)

```
if ($node instanceof Exit_) {
  return new StaticCall(new Name('BreakException'), 'exit');
}
```

zonuexe:
That's right.  Since it is difficult to process all expressions, we have dealt with only binary operations that are realistic use cases.

But your idea is smarter and far more rational.

zonuexe:
@bobthecow Thanks for reviewing!

I implemented as your suggested.  But, method name is `exit_()` because [keyword is invalid method name in PHP5](http://php.net/manual/en/reserved.keywords.php).

bobthecow:
Awesome. Thank you!

bobthecow:
I'm not super happy about `exit_`, but I'll merge and we can think about what else to call it 😛

zonuexe:
Thanks!

PR #380, opened 2017-05-28
Timer command for function time profiling

bobthecow:
Extracted from @Markcial's #138. Rebased and updated to work with the latest 0.9.x-dev stuff. Uses the new code argument input type (like `sudo` does).

Depends on #375 and #379.

### DIFF
--- a/src/Psy/Command/TimeitCommand.php
+++ b/src/Psy/Command/TimeitCommand.php
@@ -3,10 +3,11 @@
 namespace Psy\Command;
 
 use Psy\Configuration;
+use Psy\Input\CodeArgument;
 use Psy\Shell;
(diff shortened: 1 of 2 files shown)


src/Psy/Command/TimeitCommand.php
```diff
@@ -0,0 +1,66 @@
...
+
+use Psy\Configuration;
+use Psy\Input\CodeArgument;
+use Psy\Shell;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Class TimeitCommand.
```
GrahamCampbell: Not sure how useful this phpdoc is tbh. ;)

PR #389, opened 2017-06-23
Fixed slow composer performance on travis

GrahamCampbell:
This no-longer counts towards API rate limits, so is fine.


GrahamCampbell:
Takes 10-15 seconds of composer time.

bobthecow:
Thanks!

PR #390, opened 2017-06-23
Enable verbose mode on PHPUnit on Travis

GrahamCampbell:
This means that the details of the tests skipped are printed.


bobthecow:
thanks.

PR #391, opened 2017-06-24
Add Doc support for constants.

### LINKED ISSUE #384: Doc support for constants
For example:
`doc PHP_VERSION_ID`;
`=> The constant is used as an identifier for the php version...`

RyanNerd:
Closes #384

This follows similar conventions to the existing `doc` process:

* To allow for multi-language support the constants are maintained in its own SQLite database named `php_constants.sqlite`. Since system constants are not _atomically_ generated as part of PHP's manual builder process these have been added by hand to the database (English only right now).

* The 'doc keyword' command will now check if the keyword is a system constant (e.g. PHP_VERSION)
or a "magic" constant (e.g. __DIR__) if either of these is true then the documentation will be
read from the `php_constants` database and presented as output similar to how language constructs or innate functions are output.

Note: Because of the lack of reflection for these constants the output makes limited use of the reflection templating engine. I tried to "hack" the reflection helpers (like what is being done for language constructs such as `echo`) but this only proved an exercise in frustration.


bobthecow:
Aahhh! Sorry for the delay, I meant to respond to this but got behind on all the open source things :)

Thank you so much for the PR. I think the documentation generation can be folded up into the manual updates I'm doing. It's totally possible to pull this from the manual source itself and automate all the things. I'll look into that.

RyanNerd:
@bobthecow NP. Let me know if the manual updates you are performing resolves this issue and I will close the PR as unmerged.

RyanNerd:
Ping @bobthecow

bobthecow:
Sorry for the huge delay. v1.3.0 manual files now include all the constants (class constants, magic constants, normal constants).

I just pushed a change to `develop` to add support to the `doc` command. So as of the next release, it'll show documentation for constants if you've got a new enough manual db :)

PR #395, opened 2017-07-07
Add CodeCleaner\FunctionContextPass

### LINKED ISSUE #394: No support for goto
When using:
```
Psy Shell v0.8.9 (PHP 7.1.7 — cli) by Justin Hilemanan
How are you doing sir? c:
>>> label: echo 'Hello World';
Hello World⏎��
>>> goto label;
PHP Fatal error:  'goto' to undefined label 'label' in C:\Users\Jens\AppData\Roaming\Composer\vendor\psy\psysh\src\Psy\ExecutionLoop\Loop.php(90) : eval()'d code o
n line 1

Fatal error: 'goto' to undefined label 'label' in C:\Users\Jens\AppData\Roaming\Composer\vendor\psy\psysh\src\Psy\ExecutionLoop\Loop.php(90) : eval()'d code on lin
e 1
```

I get a fatal error :/

btw, is there any issue where control characters aren't ignored?

zonuexe:
Fix #394

This Pass handles `yield`, `:` (label declaration) and `goto` statement outside function context.

### DIFF
--- a/src/Psy/CodeCleaner/FunctionContextPass.php
+++ b/src/Psy/CodeCleaner/FunctionContextPass.php
@@ -50,8 +50,6 @@ class FunctionContextPass extends CodeCleanerPass
     }
 
     /**
-     * Converts exit calls to BreakExceptions.
-     *
      * @param \PhpParser\Node $node
      */
     public function leaveNode(Node $node)
--- a/src/Psy/CodeCleaner/FunctionContextPass.php
+++ b/src/Psy/CodeCleaner/FunctionContextPass.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of Psy Shell.
+ *
+ * (c) 2012-2017 Justin Hileman
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy\CodeCleaner;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Yield_;
+use PhpParser\Node\FunctionLike;
+use PhpParser\Node\Stmt\Goto_;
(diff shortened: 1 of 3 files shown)


bobthecow:
Thanks for the PR!

The `yield` handling is perfect. I think the `goto` / label handling is a bit heavy handed. for example, this is perfectly valid code:

```php
if (true) {
  start:
  echo "wat";
  if (rand(0, 1)) {
    goto start;
  }
}
```

It might make sense to do something like we do for "valid class" and "valid method" validation, where we keep track of things that were defined on the way in, then error when something does a `goto` to a label that wasn't defined during this eval?

zonuexe:
@bobthecow 
Thank you for reviewing!  I removed handling about Label from this PR.
I will send about this as another PR.

src/Psy/CodeCleaner/FunctionContextPass.php
```diff
@@ -0,0 +1,63 @@
...
+
+        // It causes fatal error.
+        if ($node instanceof Yield_) {
+            $msg = 'The "yield" expression can only be used inside a function';
+            throw new FatalErrorException($msg, 0, E_ERROR, null, $node->getLine());
+        }
+    }
+
+    /**
+     * Converts exit calls to BreakExceptions.
```
bobthecow: ☝️

zonuexe: Sorry.  I removed it.

bobthecow:
Thanks! This has been merged into the `develop` branch and will go out with the next release.

zonuexe:
I also made a mistake in choosing the merge base. Thank you!

bobthecow:
:+1:

PR #397, opened 2017-07-13
Impl LabelContextPass

zonuexe:
refs #394

This Pass simply detects invalid goto and unreachable labels.
This is a simple check, so it will overlook false negatives.
Probably not false positives. (Except `eval` 🙃  )

### DIFF
--- a/src/Psy/CodeCleaner.php
+++ b/src/Psy/CodeCleaner.php
@@ -24,6 +24,7 @@ use Psy\CodeCleaner\FunctionContextPass;
 use Psy\CodeCleaner\FunctionReturnInWriteContextPass;
 use Psy\CodeCleaner\ImplicitReturnPass;
 use Psy\CodeCleaner\InstanceOfPass;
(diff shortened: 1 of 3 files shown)


bobthecow:
I don't _think_ we need to complain about unreachable labels. `php -a` doesn't, so they're valid PHP code. And PsySH shouldn't ever reject valid PHP code. When in doubt, let the user break their session rather than rejecting valid code :)

In this case, someone might be pasting a snippet of code from a tutorial, or might be running the same code with or without a `goto` at the end, or any number of valid reasons to have a label they never reach.

```php
if (true) {
  hello:
  start:
  echo "wat";
  if (rand(0, 1)) {
    goto start;
  }
}
```

bobthecow:
@zonuexe poke ;)

bobthecow:
I've rebased and merged this, minus the assertions about unreachable labels (as mentioned in https://github.com/bobthecow/psysh/pull/397#issuecomment-315564542). It'll go in the next release!

PR #401, opened 2017-07-19
Disable phar read only

dennisdegreef:


dennisdegreef:
Rebased this and PR to develop branch

bobthecow:
It used to do this, but it makes TravisCI hang indefinitely, which breaks automatic releases :(

https://github.com/bobthecow/psysh/commit/532266b57a8a16680f3567e9118c6a7ee1f751da#diff-acb3f5be54feac3ac5f98d46a1863ca0

dennisdegreef:
wat.... :(:(:(

bobthecow:
Yeah :-/

dennisdegreef:
I was able to reproduce a infinite build locally, and the following commit fixed that behaviour, let's see what Travis thinks of this.

bobthecow:
Okay, we'll give it a shot on the next release. Thanks for digging into it :)

Out of curiosity, how did you reproduce it?

dennisdegreef:
I tried playing around with different shebangs or specifying ini settings to php -d, and I had the same results on my linux machine. Using strace, it looked like the shebang interprets the option in the shebang as script to execute, and it seems to hang in an infinite loop there. Not exactly sure what goes wrong there.

```
$ strace -ff -s 65535 ./bin/build-phar 2>&1 | grep execve
execve("./bin/build-phar", ["./bin/build-phar"], [/* 62 vars */]) = 0
execve("./bin/build-phar", ["./bin/build-phar"], [/* 63 vars */]) = 0
execve("./bin/build-phar", ["./bin/build-phar"], [/* 63 vars */]) = 0
execve("./bin/build-phar", ["./bin/build-phar"], [/* 63 vars */]) = 0
```

bobthecow:
Huh. I wonder if it's a difference in `/usr/bin/env`? Does it reproduce if you use the path to PHP directly?

dennisdegreef:
`#!/usr/bin/php -d 'phar.readonly=0'` doesn't seem to detect the option.

```
$ strace -ff ./bin/build-phar 2>&1 | grep execve
execve("./bin/build-phar", ["./bin/build-phar"], [/* 60 vars */]) = 0
$ ./bin/build-phar
PHP Fatal error:  Uncaught UnexpectedValueException: creating archive "psysh.phar" disabled by the php.ini setting phar.readonly in /src/Psy/Compiler.php:34
```

bobthecow:
Huh. Oh well :)

PR #404, opened 2017-07-25
Be able to force indexes to be shown, also for numeric arrays

dennisdegreef:
We use psysh a lot with numerically indexed arrays, where we would love to be able to see a value with it's corrosponding index number.

I dislike having to pass the value two layers deep, if you have any better way in mind, please let me know.

### DIFF
--- a/src/Psy/Configuration.php
+++ b/src/Psy/Configuration.php
@@ -47,6 +47,7 @@ class Configuration
         'defaultIncludes',
         'eraseDuplicates',
         'errorLoggingLevel',
+        'forceArrayIndexes',
         'historySize',
         'loop',
         'manualDbFile',
@@ -89,6 +90,7 @@ class Configuration
     private $colorMode;
     private $updateCheck;
     private $startupMessage;
+    private $forceArrayIndexes = false;
 
     // services
     private $readline;
@@ -1074,7 +1076,7 @@ class Configuration
     public function getPresenter()
     {
         if (!isset($this->presenter)) {
-            $this->presenter = new Presenter($this->getOutput()->getFormatter());
+            $this->presenter = new Presenter($this->getOutput()->getFormatter(), $this->getForceArrayIndexes());
         }
 
         return $this->presenter;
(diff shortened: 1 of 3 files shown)


bobthecow:
Thanks for the PR! I'd love to understand this a bit more, and see whether we can get where you need to be without a config option :)

Can you give me an example of what you're currently seeing, and what you'd like to see?

How would you feel about showing numeric array indexes only for `dump` but not for return values?

dennisdegreef:
Ofcourse, backstory is that we have used a custom-written REPL for years, and are switching to PsySh now. So a lot of our devs are used to certain behaviour, that we try to integrate.

```
>>> $repository = new FooRepository();
=> FooRepository {#171}
>>> $repository->findAll()
=> [
     Foo {#167},
     Foo {#169},
     Foo {#186},
   ]
```
Where we sometimes use these kind of methods for exploratory purposes, and we can use a certain offset of the array hardcoded in the REPL.
```
>>> $repository = new FooRepository();
=> FooRepository {#171}
>>> $repository->findAll()
=> [
     0 => Foo {#167},
     1 => Foo {#169},
     2 => Foo {#186},
   ]
>>> $repository->findAll()[1]->blah()
=> "baz"
```

dennisdegreef:
Is there anything more you'd like to know? Let me know how I can help out getting this merged :+1:

src/Psy/Configuration.php
```diff
@@ -1251,4 +1253,24 @@ public function getStartupMessage()
     {
         return $this->startupMessage;
     }
+
+    /**
+     * Get the force array indexes.
+     *
+     * @return bool
+     */
+    public function getForceArrayIndexes()
```
bobthecow: We're not 100% consistent, but most Boolean getters in Configuration don't use `get` prefix. Let's call this method `forceArrayIndexes`.

bobthecow:
Do you mind rebasing so I can merge?

dennisdegreef:
Rebased :+1:

bobthecow:
Awesome! Thanks again.

Issue #254, opened 2015-11-04
Tab completion breaks paste


tgr:
Pasting this:
```
array(
		1
);
```
(middle line indented with two tabs) into the shell (Psy Shell v0.6.0-dev (PHP 5.5.9-1ubuntu4.11 — cli)) gives
```
>>> 
Display all 3299 possibilities? (y or n) 
>>> array(
... 
Display all 3299 possibilities? (y or n) 
... 1
... );
=> [
     1
   ]

```
Somehow it should detect that the content is being pasted and ignore tabs.

For longer text, sometimes pasting is interrupted by the "Display... " line and the rest of the pasted input does not appear at all.

bobthecow:
This is an issue with readline and libedit (the libraries used to get user input by almost every command line thing ever, including PsySH) and I don't know that there's a way to fix it other than disabling tab completion :(

http://superuser.com/questions/926028/completion-when-pasting-bash-command-beginning-with-tabs

tgr:
You could try to disable autocompletion when on a line which only contains tab characters; it is [apparently possible](https://news.ycombinator.com/item?id=5660401) with the Python readline bindings at least. Or simple replace tabs with spaces. Or detect the number of characters buffered but not transmitted to readline yet, and guess whether it's a paster based on that.

bobthecow:
As far as I can tell, it's not possible with the PHP readline bindings. The tab characters seem to trigger tab completion immediately, before I have access to them on the readline buffer :-/

krageon:
It actually looks like the results returned here are from a bunch of psysh Matchers - the T_OPEN_TAG is a legitimate hasMatched for (for example) `FunctionsMatcher`, `ClassNamesMatcher` and `ConstantsMatcher`. The potential matches include things like `echo`, `empty`, etc.

This isn't a case of readline being difficult. If it was, you would be able to set `rl_completion_over` (or `completion_over` in case you are setting it using `readline_info`) to a value other than `0`.

krageon:
I made a pull request to fix this.

bobthecow:
Oh! I forgot to update this diff. This is fixed as of v0.8.11 (the readline way) if you use readline (not libedit) and a reasonably modern terminal app. Enable `useBracketedPaste` in your psysh config and enjoy :)

https://github.com/bobthecow/psysh/releases/tag/v0.8.11

krageon:
[...]

PR #410, opened 2017-08-30
Do not autocomplete if only whitespace precedes cursor

krageon:
If we are on the first character of a line and the input buffer is empty, return a tab character when autocompleting.

This should fix https://github.com/bobthecow/psysh/issues/254

### DIFF
--- a/src/Psy/TabCompletion/AutoCompleter.php
+++ b/src/Psy/TabCompletion/AutoCompleter.php
@@ -45,12 +45,12 @@ class AutoCompleter
      * Handle readline completion.
      *
      * @param string $input Readline current word
-     * @param int    $index Current word index
-     * @param array  $info  readline_info() data
-     *
+     * @param int $start Readline start match
+     * @param int $end Readline end match
+     * @param array $info readline_info() data
      * @return array
      */
-    public function processCallback($input, $index, $info = array())
+    public function processCallback($input, $start, $end, $info = array())
     {
         // Some (Windows?) systems provide incomplete `readline_info`, so let's
         // try to work around it.
@@ -60,6 +60,10 @@ class AutoCompleter
         }
         if ($line === '' && $input !== '') {
             $line = $input;
+        }
+
+        if ($this->shouldInsertOnlyTab($line, $start, $end)) {
+            return ["\t"];
         }
 
         $tokens = token_get_all('<?php ' . $line);
@@ -82,18 +86,32 @@ class AutoCompleter
     }
 
     /**
+     * @param string $line Readline current line
+     * @param int $start The start of the readline input
+     * @param int $end The end of the readline input
+     * @return bool true if the characters before $index are only whitespace
+     * @internal param int $index Current word index
+     */
+    private function shouldInsertOnlyTab($line, $start, $end)
+    {
+        $precedingInput = substr($line, $start, $end-$start);
+
+        return empty(trim($precedingInput));
+    }
+
+    /**
      * The readline_completion_function callback handler.
      *
      * @see processCallback
      *
-     * @param $input
-     * @param $index
-     *
+     * @param string $input
+     * @param int $start
+     * @param int $end
      * @return array
      */
-    public function callback($input, $index)
+    public function callback($input, $start, $end)
     {
-        return $this->processCallback($input, $index, readline_info());
+        return $this->processCallback($input, $start, $end, readline_info());
     }
 
     /**
--- a/src/Psy/TabCompletion/AutoCompleter.php
+++ b/src/Psy/TabCompletion/AutoCompleter.php
@@ -45,13 +45,12 @@ class AutoCompleter
      * Handle readline completion.
      *
      * @param string $input Readline current word
-     * @param int    $start Readline start match
-     * @param int    $end   Readline end match
+     * @param int    $index Current word index
      * @param array  $info  readline_info() data
      *
      * @return array
      */
-    public function processCallback($input, $start, $end, $info = array())
+    public function processCallback($input, $index, $info = array())
     {
         // Some (Windows?) systems provide incomplete `readline_info`, so let's
         // try to work around it.
@@ -63,7 +62,7 @@ class AutoCompleter
             $line = $input;
         }
 
-        if ($this->shouldInsertOnlyTab($line, $start, $end)) {
+        if ($this->shouldInsertOnlyTab($line, $info['point'])) {
             return array("\t");
         }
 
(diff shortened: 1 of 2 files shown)


krageon:
Travis will probably fail on old php versions as I run only 7.1 locally, I'll fix that after I have lunch.

src/Psy/TabCompletion/AutoCompleter.php
```diff
@@ -45,12 +45,13 @@ public function activate()
...
      * @param string $input Readline current word
-     * @param int    $index Current word index
+     * @param int    $start Readline start match
+     * @param int    $end   Readline end match
      * @param array  $info  readline_info() data
      *
      * @return array
      */
-    public function processCallback($input, $index, $info = array())
+    public function processCallback($input, $start, $end, $info = array())
```
bobthecow: unfortunately this is a public method, so we won't be able to break backwards compatibility. right now.

but! `$info['end']` is supposed to be a thing already, and we're using it below. we could work the `$end` param from the callback into the `$info` array to augment and/or replace that one, or we could just use that one?

or we could add the `$end` param to the end here, but that's just a bit awkward.

or we could rename this method and add a deprecated stub of a method called `processCallback` that still takes the old arguments. i think i like this third option best :)

src/Psy/TabCompletion/AutoCompleter.php
```diff
@@ -82,18 +87,37 @@ public function processCallback($input, $index, $info = array())
...
+     *
+     * @internal param int $index Current word index
+     */
+    private function shouldInsertOnlyTab($line, $start, $end)
+    {
+        $precedingInput = substr($line, 0, $end);
+
+        $trimmedInput = trim($precedingInput);
+
+        return empty($trimmedInput);
```
bobthecow: don't use empty, use `=== ''`. because php.

```php
>>> empty('0')
=> true
```

src/Psy/TabCompletion/AutoCompleter.php
```diff
@@ -82,18 +87,37 @@ public function processCallback($input, $index, $info = array())
...
-     * @param $input
-     * @param $index
+     * @param string $input
+     * @param int    $start
+     * @param int    $end
      *
      * @return array
      */
-    public function callback($input, $index)
+    public function callback($input, $start, $end)
```
bobthecow: do all the possible combinations of readline and libedit and php versions all provide `$end` for this callback? documentation is sadly lacking here :(

if not, should we fall back to the end in `readline_info` if it's available?

krageon: I looked here: https://github.com/php/php-src/blob/master/ext/readline/readline.c#L490

And I saw that `params` always contains `int start` and `int end`.

For editline / libedit, editline/readline.h is included. This has `typedef char	**rl_completion_func_t(const char *, int, int);` and `extern rl_completion_func_t *rl_attempted_completion_function;`

in editline's readline.c I see this:
```
/*
 * complete word at current point
 */
/* ARGSUSED */
int
rl_complete(int ignore __attribute__((__unused__)), int invoking_key)
{
	static ct_buffer_t wbreak_conv, sprefix_conv;
	char *breakchars;

	if (h == NULL || e == NULL)
		rl_initialize();

	if (rl_inhibit_completion) {
		char arr[2];
		arr[0] = (char)invoking_key;
		arr[1] = '\0';
		el_insertstr(e, arr);
		return CC_REFRESH;
	}

	if (rl_completion_word_break_hook != NULL)
		breakchars = (*rl_completion_word_break_hook)();
	else
		breakchars = rl_basic_word_break_characters;

	_rl_update_pos();

	/* Just look at how many global variables modify this operation! */
	return fn_complete(e,
	    (rl_compentry_func_t *)rl_completion_entry_function,
	    rl_attempted_completion_function,
	    ct_decode_string(rl_basic_word_break_characters, &wbreak_conv),
	    ct_decode_string(breakchars, &sprefix_conv),
	    _rl_completion_append_character_function,
	    (size_t)rl_completion_query_items,
	    &rl_completion_type, &rl_attempted_completion_over,
	    &rl_point, &rl_end);


}
```

Where `rl_point` and `rl_end` are updated as follows:
```
static void
_rl_update_pos(void)
{
	const LineInfo *li = el_line(e);

	rl_point = (int)(li->cursor - li->buffer);
	rl_end = (int)(li->lastchar - li->buffer);
}
```

Which leads me to believe that it does the same thing.

Even if that is in question, from the `readline.c` file in php-src, we can see that rl_end (the value passed to the callback as end) and the `end` key output by `readline_info` are the same:

https://github.com/php/php-src/blob/master/ext/readline/readline.c#L257

krageon: However, having dived into it this far, I cannot help but wonder how this works - I had interpreted `end` differently than what the documentation suggests it does, and I saw it work in that way as well. The variable I actually need is `rl_point`, which is always in `readline_info` but not passed as a callback argument.

src/Psy/TabCompletion/AutoCompleter.php
krageon: It's actually the function passed to `readline_completion_function` that needs the change, which is also public. Refactoring that in a non-breaking manner gives me a large headache, so I reverted it back to the old (and admittedly broken) way. I use `point` now, which is more in line with what should happen anyway.

krageon:
The issue that I was fixing is already fixed, so I'm closing this PR.

bobthecow:
I think this is still valuable for people who don't have a PHP or terminal or readline version that supports bracketed paste… IIRC PHP with libedit doesn't?

krageon:
Ah. I'm actually experimenting with some changes that will un-break hoa
console, they would probably also benefit from this. Alright then, I'll
polish this up (probably Monday morning).

krageon:
That's a good point. I'm also working on something that makes hoa console work properly, it's probably also handy there.

krageon:
There. That should be the tests fixed :)

src/Psy/TabCompletion/AutoCompleter.php
```diff
@@ -62,6 +62,10 @@ public function processCallback($input, $index, $info = array())
             $line = $input;
         }
 
+        if ($this->shouldInsertOnlyTab($line, $info['point'])) {
+            return array("\t");
```
bobthecow: because of the magic space inserted after tab completion, this inserts five spaces for the first tab stop, then nine spaces for each additional tab stop. why don't we hijack this and return three spaces? then it's always four space indents :)

krageon: It inserts a tab character, not spaces. I'm definitely not a fan of replacing it with space characters, as those are different (no matter what you think of tabs vs spaces) and substituting one for the other silently can never be what you want.

The tab + space behaviour is sort of vexing, but it's not something I can realistically solve in a way that doesn't break everything in environments that do work properly (for example ones that have the path from https://github.com/php/php-src/pull/2720 applied). 

What I *can* probably do is `return array("\t" . chr(8));`. `chr(8)` is the backspace character, which should work.

krageon: I could add in a setting for this and apply it to every autocompletion, thinking about it. Until the patch I mentioned works that's a sort of viable workaround to get autocompletion as you'd expect from an IDE.

krageon: Wait no, that won't work. The space is appended after the completion - so a backspace that is inserted will actually just remove the tab character entirely.

bobthecow: i understand that tabs and spaces are different. but practically, one looks like you'd expect, and the other looks really bad. so ¯\_(ツ)_/¯

GrahamCampbell:
What's the status here?

bobthecow:
This should be resolved by the new experimental readline's tab completion!

Issue #392, opened 2017-06-27
Add support for editing current command line in external editor


hackel:
Often I find myself editing long, multi-line commands, and it would be really convenient to be able to just press a key and edit it in my editor (vim).  Ideally this would work just like the edit-and-execute-command in bash (ctrl-x ctrl-e).  Since psysh is already using readline (as is bash), hopefully it won't be difficult to port.  http://git.savannah.gnu.org/cgit/bash.git/tree/bashline.c#n944

bobthecow:
Unfortunately I don't think we have access to readline in that way… the PHP implementations of readline are almost an afterthought :(

krageon:
I can look into adding the required functions as a patch to the php readline plugin.

krageon:
The referenced PR adds editing external files inline using whatever EDITOR your environment has. If/when I see that the PHP maintainers are accepting of changes to the readline extension, I will see if I can get some hotkey support.

bobthecow:
I'm going to close this for now. Feel free to reopen if/when PHP readline support is improved enough to do it with a hotkey :)

PR #411, opened 2017-08-30
edit external files and execute them inside psysh

krageon:
Adds partial support for https://github.com/bobthecow/psysh/issues/392 pending a patch to the readline extension for php that would allow me to set arbitrary hotkeys using readline.

### DIFF
--- a/src/Psy/Command/EditCommand.php
+++ b/src/Psy/Command/EditCommand.php
@@ -24,7 +24,7 @@ class EditCommand extends Command
         $filePath = $input->getArgument('file');
 
         if ($filePath === null) {
-            $filePath = tempnam(sys_get_temp_dir(), 'psysh');
+            $filePath = tempnam(sys_get_temp_dir(), 'psysh-edit-command');
         }
 
         $filePath = escapeshellarg($filePath);
--- a/src/Psy/Command/EditCommand.php
+++ b/src/Psy/Command/EditCommand.php
@@ -4,6 +4,7 @@ namespace Psy\Command;
 
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class EditCommand extends Command
@@ -16,6 +17,13 @@ class EditCommand extends Command
             ->setName('edit')
             ->setDefinition(array(
                 new InputArgument('file', InputArgument::OPTIONAL, 'The file to open for editing. If this is not given, edits a temporary file.', null),
+                new InputOption(
+                    'exec',
+                    'e',
+                    InputOption::VALUE_OPTIONAL,
+                    'Whether or not to execute the file content after editing. Defaults to true when no file was provided, defaults to false if a file was provided',
+                    null
+                ),
             ))
             ->setDescription('Open an external editor. Afterwards, get produced code in input buffer.')
             ->setHelp('Set the EDITOR environment variable to something you\'d like to use.');
@@ -27,7 +35,12 @@ class EditCommand extends Command
 
         if ($filePath === null) {
             $filePath = tempnam($this->temporaryDirectory, 'psysh-edit-command');
+            $execute = $input->getOption('exec') ?: 'true';
+        } else {
+            $execute = $input->getOption('exec') ?: 'false';
         }
+
+        $execute = $execute === 'true';
 
         $escapedFilePath = escapeshellarg($filePath);
 
@@ -41,7 +54,9 @@ class EditCommand extends Command
             @unlink($filePath);
         }
 
-        $this->getApplication()->addInput($editedContent);
+        if ($execute) {
+            $this->getApplication()->addInput($editedContent);
+        }
     }
 
     public function setTemporaryDirectory($tmpDir)
--- a/src/Psy/Command/EditCommand.php
+++ b/src/Psy/Command/EditCommand.php
@@ -33,11 +33,11 @@ class EditCommand extends Command
     {
         $filePath = $input->getArgument('file');
 
+        $execute = $input->getOption('exec') ?: 'false';
+
         if ($filePath === null) {
             $filePath = tempnam($this->temporaryDirectory, 'psysh-edit-command');
             $execute = $input->getOption('exec') ?: 'true';
-        } else {
-            $execute = $input->getOption('exec') ?: 'false';
         }
 
         $execute = $execute === 'true';
--- a/src/Psy/Command/EditCommand.php
+++ b/src/Psy/Command/EditCommand.php
@@ -6,13 +6,12 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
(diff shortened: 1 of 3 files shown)


src/Psy/Command/ExternalEditorCommand.php
```diff
@@ -0,0 +1,43 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $filePath = $input->getArgument('file');
+
+        if ($filePath === null) {
+            $filePath = tempnam(sys_get_temp_dir(), 'psysh');
+        }
+
+        $pipes = array();
+        $proc = proc_open((getenv('EDITOR') ?: 'nano') . " {$filePath}", array(STDIN, STDOUT, STDERR), $pipes);
```
the-eater: use [`escapeshellarg`](http://php.net/escapeshellarg)

krageon: good point!

src/Psy/Command/ExternalEditorCommand.php
```diff
@@ -0,0 +1,45 @@
...
+            ->setDescription('Open an external editor. Afterwards, get produced code in input buffer.')
+            ->setHelp('Set the EDITOR environment variable to something you\'d like to use.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $filePath = $input->getArgument('file');
+
+        if ($filePath === null) {
+            $filePath = tempnam(sys_get_temp_dir(), 'psysh');
```
bobthecow: Please pass in the "runtime dir" from Shell::config in the constructor and use that rather than `sys_get_temp_dir`. It's effectively the same, but can be overridden via XDG for systems that support it.

src/Psy/Command/ExternalEditorCommand.php
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        $filePath = escapeshellarg($filePath);
+
+        $pipes = array();
+        $proc = proc_open((getenv('EDITOR') ?: 'nano') . " {$filePath}", array(STDIN, STDOUT, STDERR), $pipes);
+        proc_close($proc);
+
+        $editedContent = file_get_contents($filePath);
```
bobthecow: Your `escapeshellarg` change broke `file_get_contents`.

src/Psy/Command/ExternalEditorCommand.php
bobthecow: Also, please name the tempfile with `externaledit` or similar as the prefix; `psysh` is too generic.

bobthecow:
Thank you! A see the small fixes inline :)

Also:

* Please change the merge target to the `develop` branch.
* It would be nice to support `edit $variable`, especially `edit $__file` to edit the magic shell variable supplied by `doc ClassName`, `show functionName`, etc. See `ThrowUpCommand` for an example of a `ContextAware` command that pulls variables out of the shell context.
* Iff a file argument is supplied, we probably don't want to edit-and-execute, at least by default. Maybe add a flag for `--eval` or `--exec` or something, and have it default to `true` if no file arg is supplied, and false if one is?

src/Psy/Command/ExternalEditorCommand.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ExternalEditorCommand extends Command
+{
+    protected function configure()
+    {
+        $this
+            ->setName('externaledit')
```
bobthecow: I'd call the whole thing `edit` and skip the `externaledit` business entirely :)

krageon:
Thanks for the feedback! I'll work on incorporating it right away.

krageon:
Your review comments went all weird because I rebased the feature branches onto your develop branch. Apologies for that.

krageon:
Support for the `$variable` input syntax in `EditCommand` will either come later this evening or (more likely) tomorrow morning. Thanks for the input so far :)

src/Psy/Command/EditCommand.php
```diff
@@ -0,0 +1,66 @@
...
+        if ($input->getArgument('file') === null) {
+            @unlink($filePath);
+        }
+
+        if ($execute) {
+            $this->getApplication()->addInput($editedContent);
+        }
+    }
+
+    public function setTemporaryDirectory($tmpDir)
```
bobthecow: i'd prefer passing it in the constructor. this value is required for making this command work properly; there's no need to add a setter :)

src/Psy/Command/EditCommand.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Psy\Command;
+
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class EditCommand extends Command
+{
+    private $temporaryDirectory = '';
```
bobthecow: call it `runtimeDir` for consistency with XDG and everywhere else we call it that?

src/Psy/Command/EditCommand.php
```diff
@@ -0,0 +1,66 @@
...
+    protected function configure()
+    {
+        $this
+            ->setName('edit')
+            ->setDefinition(array(
+                new InputArgument('file', InputArgument::OPTIONAL, 'The file to open for editing. If this is not given, edits a temporary file.', null),
+                new InputOption(
+                    'exec',
+                    'e',
+                    InputOption::VALUE_OPTIONAL,
```
bobthecow: should be `VALUE_NONE`; this is a boolean flag.

src/Psy/Command/EditCommand.php
```diff
@@ -0,0 +1,171 @@
...
+     * @return bool
+     */
+    private function shouldExecuteFile($execOption, $noExecOption, $filePath)
+    {
+        if ($execOption) {
+            return true;
+        }
+
+        if ($noExecOption) {
+            return false;
```
the-eater: Don't you want the `no-exec` to override `exec`? as `no-exec` is always safer?

the-eater: Ah I said nothing you don't allow both

krageon: Exactly like that. Using both flags makes no sense and should throw an error, which it does.

src/Psy/Command/EditCommand.php
```diff
@@ -0,0 +1,171 @@
...
+                    InputOption::VALUE_NONE,
+                    'Do not execute the file content after editing. This is the default when a file name argument is given.',
+                    null
+                ),
+            ))
+            ->setDescription('Open an external editor. Afterwards, get produced code in input buffer.')
+            ->setHelp('Set the EDITOR environment variable to something you\'d like to use.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
```
bobthecow: mind adding a `@throws` comment here?

krageon:
Of course. I should probably configure my editor to do it for me, now that
you mention it.

bobthecow:
Thank you :)

PR #412, opened 2017-08-30
complete default parameters

krageon:
When a function (either of a class, an object or just a bare function) has only default parameters, autocomplete will now complete the function call for you with those parameters substituted in.


src/Psy/TabCompletion/Matcher/ObjectMethodDefaultParametersMatcher.php
```diff
@@ -0,0 +1,95 @@
...
+
+        return array();
+    }
+
+    /**
+     * @param \ReflectionMethod $method
+     *
+     * @return string
+     */
+    private function extractParameterString($method)
```
bobthecow: please use types for arguments where possible.

src/Psy/TabCompletion/Matcher/ObjectMethodDefaultParametersMatcher.php
```diff
@@ -0,0 +1,95 @@
...
+        $parametersProcessed = array();
+
+        foreach ($parameters as $parameter) {
+            if (!$parameter->isDefaultValueAvailable()) {
+                return '';
+            }
+
+            $defaultValue = $parameter->getDefaultValue();
+            if (is_string($defaultValue)) {
+                $defaultValue = "'{$defaultValue}'";
```
bobthecow: use `var_export` here to escape the string and add quotes to it.

src/Psy/TabCompletion/Matcher/ObjectMethodDefaultParametersMatcher.php
krageon: That's a good point. I'm used to putting type hinting everywhere, but I removed it all when I realised that wouldn't work all the way back to PHP 5.3. I was a little overzealous.

bobthecow:
This one is great! There's a ton of duplicated code between the matchers, though. Can the object and class method matchers extend the function matcher? Or can they all extend an abstract function matcher or something?

krageon:
Good point. I split it out into an abstract default parameter matcher that has a function to do most of the common stuff.

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
```diff
@@ -0,0 +1,32 @@
...
+            $defaultValue = var_export($parameter->getDefaultValue(), true);
+
+            $parametersProcessed[] = "\${$parameter->getName()} = $defaultValue";
+        }
+
+        if (empty($parametersProcessed)) {
+            return array();
+        }
+
+        return array(implode(',', $parametersProcessed) . ')');
```
bobthecow: this should implode on `, ` (with a space).

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
```diff
@@ -0,0 +1,32 @@
...
+    public function getDefaultParameterCompletion(array $reflectionParameters)
+    {
+        $parametersProcessed = array();
+
+        foreach ($reflectionParameters as $parameter) {
+            if (!$parameter->isDefaultValueAvailable()) {
+                return array();
+            }
+
+            $defaultValue = var_export($parameter->getDefaultValue(), true);
```
bobthecow: After using this for a bit, it feels like `var_export` isn't 100% right here.  for example, `var_export(array())` is a bit ugly:

```
Psy\Shell::debug($vars = array (
),$boundObject = NULL)
```

Ideally, this would complete as:

```
Psy\Shell::debug($vars = [], $boundObject = NULL)
```

krageon: This is very true. I suppose I could parse the code using the AST parser and make a custom formatter for shortform everything, but that would be a lot slower than what is there now. Another option would be to throw some regex at this problem, which is probably a little faster but also a lot less easy to modify. Do you have a preference?

krageon: I just tried mocking this, but it's actually a lot less trivial than I thought initially. I'll table it for now and return to it a little later.

bobthecow: Since it's just for default values, a few simple transforms would probably be fine. The most common case where var export is wrong will likely be an empty array? We could start be special casing just that one. And maybe deal with new lines?

krageon: A filled array is also strange. With multiline strings as default parameters, you get newlines but not ones you want to replace. Then you can have multiline strings inside of arrays as default parameters, that is where I decided I didn't want regex. I spent a little bit getting a prettyprinter to work reliable but failed, but that definitely doesn't mean it's impossible - I just have to mull over what I did wrong.

bobthecow: How about this:

```php
function stringify($value) {
  if (is_array($value)) {
    $chunks = [];
    foreach ($value as $key => $val) {
      if ($key !== count($chunks)) {
        return false;
      }

      $chunks[] = stringify($val);
    }

    return '[' . implode(', ', $chunks) . ']';
  }

  // Use json_encode because it's a close enough approximation of what we're
  // looking for with strings, numbers, booleans and null...
  return json_encode($value);
}
```

If we can encode this as a "clean" array, do that, as above. but if `stringify` returns `false`, bail completely and don't complete default parameters.

Using `json_encode` rather than `var_export` solves for `\n` in strings, and special casing all arrays prevents it from generating invalid PHP.

As a bonus, it'll give us `true`, `false`, and `null` rather than the more shouty `TRUE`, `FALSE` and `NULL` that `var_exports` returns.

As an extra bonus, it doesn't use output buffering internally so we can use it wherever we want :)

krageon: That's an interesting solution. I'll test that out (probably in the morning), thanks!

bobthecow: poke :)

krageon: Apologies, I didn't forget about you :) I got sidetracked by something else. I'll look at this today assuming nothing explodes.

bobthecow: No worries. Just wanted to make sure it didn't get dropped.

krageon:
Definitely not - I use this quite often. I'm not 100% happy with using
json_encode though. It'd have my preference if the solution was as close to
being exactly what php sees as possible. To that end, I'm looking for a way
to get the code so I can get the AST from php-parser. That will allow me to
make a custom pretty printer, which will allow fixing basically every
presentation issue.

krageon:
In the interest of making something that works now instead of ten weeks, your solution is definitely the superior one :)

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
```diff
@@ -0,0 +1,59 @@
...
+    private function valueToShortString($value)
+    {
+        if (!is_array($value)) {
+            return json_encode($value);
+        }
+
+        $chunks = '';
+
+        foreach ($value as $key => $item) {
+            $keyString = $this->valueToShortString($key);
```
bobthecow: it would be nice to leave off the keyString if the array has consecutive zero-based numeric indices.

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
```diff
@@ -0,0 +1,59 @@
...
+            $itemString = $this->valueToShortString($item);
+
+            $chunks .= "{$keyString} => {$itemString}, ";
```
bobthecow: this will leave a trailing `, ` which is a bit awkward.

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
bobthecow: ☝️ still this :)

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
krageon: Good point. I'll `implode` (with a space this time haha)

krageon:
I'll fix styleci and test it in the morning :)

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
```diff
@@ -0,0 +1,71 @@
...
+            if ($allSequential === true) {
+                $allSequential = is_numeric($key) && ($key === ($lastKey + 1));
+                $lastKey = $key;
+            }
+
+            $keyString = $this->valueToShortString($key);
+            $itemString = $this->valueToShortString($item);
+
+            $chunks[] = "{$keyString} => {$itemString}";
+            $chunksSequential[] = "{$itemString}";
```
bobthecow: unnecessary interpolation here.

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        $chunksToImplode = $allSequential ? $chunksSequential : $chunks;
+
+        return '[ ' . implode(', ', $chunksToImplode) . ' ]';
```
bobthecow: let's do it with no space inside the brackets — `[1, 2, 3]` instead of `[ 1, 2, 3 ]`.

bobthecow:
Looking good. Just two quick fixes!

src/Psy/TabCompletion/Matcher/AbstractDefaultParametersMatcher.php
```diff
@@ -0,0 +1,71 @@
...
+
+        $chunks = array();
+        $chunksSequential = array();
+
+        $lastKey = -1;
+        $allSequential = true;
+
+        foreach ($value as $key => $item) {
+            if ($allSequential === true) {
+                $allSequential = is_numeric($key) && ($key === ($lastKey + 1));
```
bobthecow: you can skip all the lastKey business by checking the length of chunksSequential…

```
$allSequential = $allSequential && is_numeric($key) && $key === count($chunksSequential);
```

krageon: That's neat. I'll fix it

bobthecow:
🎉 ‼️

PR #413, opened 2017-08-31
double colon suggest

krageon:
Suggest a double colon if you have typed the name of a class that has static methods.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ manual/
 dist/
 __pycache__
 .php_cs.cache
+psysh.phar
+psysh-compat.phar
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,3 @@ manual/
 dist/
 __pycache__
 .php_cs.cache
-psysh.phar
-psysh-compat.phar
-.idea/
--- a/src/Psy/Shell.php
+++ b/src/Psy/Shell.php
@@ -200,6 +200,7 @@ class Shell extends Application
                 new Matcher\VariablesMatcher(),
                 new Matcher\ConstantsMatcher(),
                 new Matcher\FunctionsMatcher(),
+                new Matcher\ClassDoubleColonMatcher(),
                 new Matcher\ClassNamesMatcher(),
                 new Matcher\ClassMethodsMatcher(),
                 new Matcher\ClassAttributesMatcher(),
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -7,3 +7,6 @@ manual/
 dist/
 __pycache__
 .php_cs.cache
+psysh.phar
+psysh-compat.phar
+.idea/
```
bobthecow: Please add this one to your global gitignore rather than leaving it as a little present in every open source project you contribute to :)

krageon: Good point - done!

krageon:
Travis fails for PHP 5.3, but it looks like that's a problem that doesn't have anything to do with this pull request.

bobthecow:
Yeah, it looks like Travis started building 5.3 with trusty instead of precise. I re-ran `develop` build and it used precise, so I'm re-running the builds for these pull requests to see if they resolve themselves.

bobthecow:
Nope. No luck. I'll try explicitly using precise for 5.3.

bobthecow:
Yep, that worked!

https://travis-ci.org/bobthecow/psysh/builds/271876437

Do you mind rebasing (this and the other branches) so we can get a green build before merging?

krageon:
I did not rebase them, as I've used these branches as a base of something else as well. This would make merging all of the eventual changes (including ones that I cannot publish yet because the required changes are not yet in the readline extension) very difficult. I *did* merge it and it works now, so that is good.

bobthecow:
With this one, I can't get it to actually complete as anything useful… If I type `Psy\Shell` and hit tab a couple of times, it adds an extraneous space after `Psy\Shell`, then it shows `Psy\Shell::` and `Shell`.

But if I actually try to complete the `Psy\Shell::`, I end up with `Psy\Shell Psy\Shell::` on my line.

krageon:
Yes, that is because out of the box your readline completes with a space after the end. The double completion you see can be fixed by this merge request: https://github.com/php/php-src/pull/2720

I have another one in the bleachers (that will not work with things as it stands, because that readline fix does not yet exist) for psysh that removes the appended space after completion. When that is merged, you can press tab to complete the class name and press tab again to complete the double colon (with a menu completion shortcut set, you can press that key to cycle through whatever options you have).

bobthecow:
If that's the case, would it make sense to hold off on this one unless/until the updates to PHP are released?

krageon:
Sure! I'll close it pending that change.

bobthecow:
Sounds good!

PR #415, opened 2017-09-05
sets php-xdg-base-dir to dev-master

krageon:
pending my PR being accepted or me adding it to packagist.

Of course, I see there are pull requests from 2014 open. I'll put hard forking it and integrating all the open pull requests on my todo list.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,8 @@
         "symfony/console": "~2.3.10|^2.4.2|~3.0",
         "symfony/var-dumper": "~2.7|~3.0",
         "nikic/php-parser": "~1.3|~2.0|~3.0",
-        "dnoegel/php-xdg-base-dir": "0.1",
-        "jakub-onderka/php-console-highlighter": "0.3.*"
+        "jakub-onderka/php-console-highlighter": "0.3.*",
+        "krageon/php-xdg-base-dir": "^0.1.0"
     },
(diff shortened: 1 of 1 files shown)


krageon:
The merge request in question: https://github.com/dnoegel/php-xdg-base-dir/pull/20

krageon:
I've changed this so the used version is set to `dev-master` of the original package. Anything lower doesn't have my changes, but my PR was accepted so using a fork is no longer necessary.

theofidry:
@bobthecow I would consider this PR (needs to be rebased but still). Despite being annoying to depend on dev-master, there is [a few changes](https://github.com/dnoegel/php-xdg-base-dir/compare/0.1...master) which would be really great. An alternative would be to get write access on the repository to manage it or a fork...

GrahamCampbell:
The upstream package doesn't seem to be maintained. I wonder if we should just bring in the stuff we need, directly into this package.

theofidry:
Let's check with the maintainer first if he is willing to gives access to his repo: https://twitter.com/tfidry/status/1201494486347632640

GrahamCampbell:
@theofidry Seems like a good plan. Ordinarily, I'd suggest this, but since It's only a single file package, it's maybe not worth the effort. We can certainly see what the package author says.

PR #417, opened 2017-09-11
Autocomplete class/object: Error handling

krageon:
Adds some extra checks and safeguards to prevent various completion-related errors from popping up:

Objects:
```
>>> $no = [];
>>> $no-><tab> // gives an error
```

Classes are basically the same deal, except it'd be the double colon operator and the errors take a bit to pop up (various incarnations of "this class does not exist")


bobthecow:
Thanks!

PR #420, opened 2017-09-18
Do not suggest super methods

jens1o:
fixes #419

PR #421, opened 2017-09-18
Do not suggest super methods

### LINKED ISSUE #419: FR: Provide no suggestions for `__construct`, `__wakeup`...
Because it is not intended to do so, but psysh still suggests its usage. :/

jens1o:
fixes #419 

#420

### DIFF
--- a/src/Psy/TabCompletion/Matcher/ObjectMethodsMatcher.php
+++ b/src/Psy/TabCompletion/Matcher/ObjectMethodsMatcher.php
@@ -50,10 +50,11 @@ class ObjectMethodsMatcher extends AbstractContextAwareMatcher
         return array_filter(
             get_class_methods($object),
             function ($var) use ($input) {
-                return
-                    AbstractMatcher::startsWith($input, $var)
+                return (
+                    AbstractMatcher::startsWith($input, $var) &&
                     // also check that we do not suggest invoking a super method(__construct, __wakeup, …)
-                    && !AbstractMatcher::startsWith('__', $var);
+                    !AbstractMatcher::startsWith('__', $var)
+                );
             }
         );
     }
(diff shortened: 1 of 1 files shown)


jens1o:
yeeeyy, it works 😃

src/Psy/TabCompletion/Matcher/ObjectMethodsMatcher.php
```diff
@@ -50,7 +50,10 @@ public function getMatches(array $tokens, array $info = array())
         return array_filter(
             get_class_methods($object),
             function ($var) use ($input) {
-                return AbstractMatcher::startsWith($input, $var);
+                return
```
bobthecow: I know it's completely arbitrary, and the linter didn't even complain, but for a multiline return I prefer to wrap it in parens (as if it were a multi line array).

jens1o: Okay, will do that this afternoon.

jens1o:
```php

                return
                    (AbstractMatcher::startsWith($input, $var)
                    // also check that we do not suggest invoking a super method(__construct, __wakeup, …)
                    && !AbstractMatcher::startsWith('__', $var));
```

Do you expect it like this?

bobthecow:
```php
return (
    AbstractMatcher::startsWith($input, $var) &&
    // also check that we do not suggest invoking a super method(__construct, __wakeup, …)
    !AbstractMatcher::startsWith('__', $var)
);
```

jens1o:
Okay, now it complies with the code style :/

bobthecow:
Gah. Sorry. 

Okay, do…

```
return foo &&
    bar;
```

?

jens1o:
No worries, just let's hope together.

bobthecow:
Thanks!

PR #425, opened 2017-09-23
Improve performance of AbstractMatcher#startsWith

jens1o:
regular expressions are always slow. :/

### DIFF
--- a/src/Psy/TabCompletion/Matcher/AbstractMatcher.php
+++ b/src/Psy/TabCompletion/Matcher/AbstractMatcher.php
@@ -113,7 +113,7 @@ abstract class AbstractMatcher
     public static function startsWith($prefix, $word)
     {
         if (empty($prefix)) {
-             return true;
+            return true;
         }
 
         return strpos($word, $prefix) === 0;
--- a/src/Psy/TabCompletion/Matcher/AbstractMatcher.php
+++ b/src/Psy/TabCompletion/Matcher/AbstractMatcher.php
@@ -112,7 +112,9 @@ abstract class AbstractMatcher
      */
     public static function startsWith($prefix, $word)
     {
-        if (empty($prefix)) return true;
+        if (empty($prefix)) {
+             return true;
+        }
 
         return strpos($word, $prefix) === 0;
     }
--- a/src/Psy/TabCompletion/Matcher/AbstractMatcher.php
+++ b/src/Psy/TabCompletion/Matcher/AbstractMatcher.php
@@ -112,7 +112,9 @@ abstract class AbstractMatcher
      */
     public static function startsWith($prefix, $word)
     {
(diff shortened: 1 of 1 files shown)


jens1o:
Okay, I broke it.

src/Psy/TabCompletion/Matcher/AbstractMatcher.php
```diff
@@ -112,7 +112,11 @@ protected function getNamespaceAndClass($tokens)
      */
     public static function startsWith($prefix, $word)
     {
-        return preg_match(sprintf('#^%s#', $prefix), $word);
+        if (empty($prefix)) {
```
bobthecow: you can't use `empty` with strings in PHP.

```php
empty("0")
```

ಠ_ಠ

jens1o: you shouldn't program in two different programming languages on two monitors at the same time.

jens1o:
@bobthecow Can you point me why that breaks there, I didn't expected to make any type of api change and I don't see why it fails :(

bobthecow:
To be honest, I'm not sold on the premise. Did you test it?

Regexes are faster than you think ;)

Your implementation is pretty quick in the best case (if the pattern is found at the start of a string) but suffers quite a bit in the average and worst case:

```
strpos (match at start):   2.1370
strpos (match at middle): 83.4229
strpos (match at end):    84.7981
strpos (no match):        84.6062
```

There are 4013 defined functions, classes, interfaces and constants when I spin up a fresh `psysh` shell. It is _extremely_ likely that most or all of the time only a couple of these will match the current input (i.e. it's usually the worst case above, not the best). So despite convincing best case performance, it's not what we should be optimizing for.

The existing regex-based implementation works pretty well across the board:

```
Regex (match at start):    3.8769
Regex (match at middle):   3.6940
Regex (match at end):      3.5827
Regex (no match):          3.7838
```

When testing a matching string (remember: least common thing this will do) it's around 80% slower than `strpos`. But when testing a string that matches at the middle or end it's _slightly faster_ than finding a match at the start because it only has to test _one character_ before it knows the pattern doesn't match!

A bonus implementation I just wrote using `substr` performs a bit better than the regex-based version:

```
substr (match at start):   2.9382
substr (match at middle):  2.9085
substr (match at end):     2.9813
substr (no match):         2.9530
```

... but honestly, not enough better to worry about. Keep in mind that the difference between these are microscopic. The numbers listed above are in seconds, but I had to do _2.5 million iterations_ just to make the tests take seconds rather than fractions of seconds that are too small to measure.

jens1o:
>To be honest, I'm not sold on the premise. Did you test it?

Yeah, and I thought that we would have more cases like the first one, but that was indeed very naive. :/

PR #437, opened 2017-11-11
PSR2 fix minor violations

keradus:
Brought to you by [PHP CS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)

### DIFF
--- a/src/Psy/Command/ReflectingCommand.php
+++ b/src/Psy/Command/ReflectingCommand.php
@@ -69,6 +69,7 @@ abstract class ReflectingCommand extends Command implements ContextAware
                     throw new RuntimeException('Unknown target: ' . $valueName);
                 }
 
+                // no break
             case preg_match(self::CLASS_OR_FUNC, $valueName, $matches):
                 return array($this->resolveName($matches[0], true), null, 0);
 
--- a/src/Psy/Command/ReflectingCommand.php
+++ b/src/Psy/Command/ReflectingCommand.php
@@ -65,11 +65,9 @@ abstract class ReflectingCommand extends Command implements ContextAware
                 // @todo maybe do something interesting with these at some point?
                 if (array_key_exists($matches[1], $GLOBALS)) {
                     throw new RuntimeException('Unable to inspect a non-object');
-                } else {
-                    throw new RuntimeException('Unknown target: ' . $valueName);
                 }
 
-                // no break
+                throw new RuntimeException('Unknown target: ' . $valueName);
             case preg_match(self::CLASS_OR_FUNC, $valueName, $matches):
                 return array($this->resolveName($matches[0], true), null, 0);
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
These violations would have already been automatically fixed. :P

keradus:
PHP CS Fixer (so style ci which relies on it) could fix it automatically indeed, yet for some reason those two were not applied to repo.

GrahamCampbell:
Take a look at develop branch. It'll get the fixes first.

src/Psy/Command/ReflectingCommand.php
```diff
@@ -69,6 +69,7 @@ protected function getTarget($valueName, $classOnly = false)
                     throw new RuntimeException('Unknown target: ' . $valueName);
                 }
 
+                // no break
```
GrahamCampbell: This comment is misleading, since we do not fall through. 👎 for adding it.

src/Psy/Shell.php
```diff
@@ -1013,7 +1013,7 @@ protected function writeVersionInfo()
         try {
             $client = $this->config->getChecker();
             if (!$client->isLatest()) {
-                $this->output->writeln(sprintf('New version is available (current: %s, latest: %s)',self::VERSION, $client->getLatest()));
+                $this->output->writeln(sprintf('New version is available (current: %s, latest: %s)', self::VERSION, $client->getLatest()));
```
GrahamCampbell: This was not fixed by StyleCI because the fixer is turned off in the configuration.

keradus: still, PSR2 violation

src/Psy/Command/ReflectingCommand.php
```diff
@@ -65,10 +65,9 @@ protected function getTarget($valueName, $classOnly = false)
                 // @todo maybe do something interesting with these at some point?
                 if (array_key_exists($matches[1], $GLOBALS)) {
                     throw new RuntimeException('Unable to inspect a non-object');
-                } else {
-                    throw new RuntimeException('Unknown target: ' . $valueName);
                 }
 
+                throw new RuntimeException('Unknown target: ' . $valueName);
```
GrahamCampbell: Bit of a hack to get PHP-CS-Fixer to work. :trollface:

keradus: else after `return`/`break`/`continue`/`throw` is always sad.

bobthecow: :+1:

src/Psy/Shell.php
bobthecow: this is `method_argument_space`, right? is that the one we disabled because it prevents lining up arguments on consecutive lines?

keradus: yes. yet, i didn't enabled the rule or apply it globally, this was the line where space was simply missing, and you are talking about cases where there were multiple spaces for aligning purposes

PR #449, opened 2017-12-24
Bound the required php version

GrahamCampbell:


bobthecow:
Is this a good idea? With PHP's track record for backwards compatibility, I imagine PsySH will keep working beyond PHP 7.x.

GrahamCampbell:
I'd recommend that version constraints are always bounded. It doesn't hurt to raise the bound later down the line if it works with 8.x, but impossible to do the opposite.

bobthecow:
My worry with that is that a project will set a psysh dependency to `0.9.*` and never update it… note that this already happens. Check packagist for v0.6.x installs… there's a steady trickle of them, _years later_.

I don't want PsySH to be the only reason one of those projects gets stuck with an old version of PHP.

Granted, I could fix (most of) this at whatever future point PHP 8.x ships by cutting a new PsySH release for every semver major and minor version we've ever shipped, which just bumps the PHP constraint. But that sounds like a lot of work for very little gain. And it only solves for projects that use semver constraints for their PsySH dependency. Hard coded versions will never pick it up without an update :(

GrahamCampbell:
Your decision, but I think the least worst option is to always bound dependencies. :)

GrahamCampbell:
Have you decided what to do here?

Taluu:
You shouldn't event consider the older versions IMO. If the guy doesn't want to upgrade, too bad for him. If he wants the new features, he should upgrade. And that's especially true for 0.x versions.

bobthecow:
@GrahamCampbell I can't quite make myself comfortable with this.

In principle, I agree with bounding dependencies. In this specific case, though, I'm less worried about PHP shipping backwards compatibility breaks than I am worried about other libraries pinning their PsySH version and not updating, sticking end users with a hard to resolve dependency.

@Taluu I agree completely, for _projects_. For _libraries_, which are used as dependencies by a lot of other projects, I've gotta err on the side of compatibility. The tighter dependencies are locked down, the greater the chance that we can put end users into a position where they can't possibly meet the version constraints of all the packages they want to use together.

PR #454, opened 2018-01-01
Shell::execute() and timeit command.

bobthecow:


src/Psy/ExecutionClosure.php
```diff
@@ -0,0 +1,103 @@
...
+                    ob_end_clean();
+                }
+
+                throw $_e;
+            }
+
+            return $_;
+        };
+
+        if (self::bindClosure()) {
```
GrahamCampbell: Confusing method name perhaps? Better to name it `shouldBindClosure`?

src/Psy/ExecutionLoop.php
```diff
@@ -21,128 +21,71 @@
...
     /**
      * Run the execution loop.
      *
      * @throws ThrowUpException if thrown by the `throw-up` command
      *
-     * @param Shell $shell
+     * @param Shell &$shell
      */
-    public function run(Shell $shell)
+    public function run(Shell &$shell)
```
GrahamCampbell: Why do we need the `&`?

src/Psy/ExecutionClosure.php
```diff
@@ -0,0 +1,103 @@
...
+ * The Psy Shell's execution scope.
+ */
+class ExecutionClosure
+{
+    const NOOP_INPUT = 'return null;';
+
+    private $closure;
+
+    /**
+     * @param Shell &$__psysh__
```
GrahamCampbell: Why the `&`. Objects are already passed by reference.

src/Psy/ExecutionClosure.php
```diff
@@ -0,0 +1,103 @@
...
+            try {
+                // Restore execution scope variables
+                extract($__psysh__->getScopeVariables(false));
+
+                // evaluate the current code buffer
+                ob_start([$__psysh__, 'writeStdout'], 1);
+
+                set_error_handler([$__psysh__, 'handleError']);
+                $_ = eval($__psysh__->onExecute($__psysh__->flushCode() ?: ExecutionClosure::NOOP_INPUT));
+                restore_error_handler();
```
GrahamCampbell: Probably want this in a finally?

src/Psy/ExecutionClosure.php
```diff
@@ -0,0 +1,103 @@
...
+
+                ob_end_flush();
+
+                $__psysh__->setScopeVariables(get_defined_vars());
+            } catch (\Throwable $_e) {
+                restore_error_handler();
```
GrahamCampbell: Are we sure we want to call this again if we already called it?

src/Psy/ExecutionClosure.php
bobthecow: :+1:

src/Psy/ExecutionClosure.php
bobthecow: I did think about this when cleaning this up. I'm not sure we do. `handleError` should only handle errors during `eval` (or as close to it as possible). If we put it in a finally, it'll handle errors in the flush/clean, as well as errors in setting scope variables. We could narrow it down to just handling errors in eval and the `ob_end_flush`, which is probably good enough…

```diff
diff --git a/src/Psy/ExecutionClosure.php b/src/Psy/ExecutionClosure.php
index 7dfd7eb..5e57d12 100644
--- a/src/Psy/ExecutionClosure.php
+++ b/src/Psy/ExecutionClosure.php
@@ -32,30 +32,20 @@ class ExecutionClosure
 
                 // evaluate the current code buffer
                 ob_start([$__psysh__, 'writeStdout'], 1);
-
                 set_error_handler([$__psysh__, 'handleError']);
                 $_ = eval($__psysh__->onExecute($__psysh__->flushCode() ?: ExecutionClosure::NOOP_INPUT));
-                restore_error_handler();
 
+                // Has to be done before the finally or it'll get cleaned rather than flushed.
                 ob_end_flush();
-
-                $__psysh__->setScopeVariables(get_defined_vars());
-            } catch (\Throwable $_e) {
+            } finally {
                 restore_error_handler();
                 if (ob_get_level() > 0) {
                     ob_end_clean();
                 }
-
-                throw $_e;
-            } catch (\Exception $_e) {
-                restore_error_handler();
-                if (ob_get_level() > 0) {
-                    ob_end_clean();
-                }
-
-                throw $_e;
             }
 
+            $__psysh__->setScopeVariables(get_defined_vars());
+
             return $_;
         };
 ```

... but it's all a moot point because `finally` is PHP 5.5+ and this version of PsySH is still targeting >= 5.4 :(

src/Psy/ExecutionClosure.php
bobthecow: Done. Thanks for catching this.

PR #458, opened 2018-01-08
Improve assertions

carusogabriel:
I've made a little improvement to our tests:
- use PHPUnit assertion methods instead of PHP functions;
- use `assertSame` (===) instead of `assertEquals` (==)


bobthecow:
Great, thank you!

PR #460, opened 2018-01-10
Fixes from PHPStan

carusogabriel:
Some fixes from [`PHPStan`](https://github.com/phpstan/phpstan) levels 0, 1 and 2 :man_technologist:

### DIFF
--- a/src/Psy/Input/FilterOptions.php
+++ b/src/Psy/Input/FilterOptions.php
@@ -11,6 +11,7 @@
 
 namespace Psy\Input;
 
+use Psy\Exception\ErrorException;
 use Psy\Exception\RuntimeException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
--- a/src/Psy/TabCompletion/Matcher/AbstractContextAwareMatcher.php
+++ b/src/Psy/TabCompletion/Matcher/AbstractContextAwareMatcher.php
@@ -44,7 +44,7 @@ abstract class AbstractContextAwareMatcher extends AbstractMatcher implements Co
     /**
      * Get a Context variable by name.
      *
-     * @param $var Variable name
+     * @param string $var Variable name
      *
      * @return mixed
      */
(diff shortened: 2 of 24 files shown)


src/Psy/Input/FilterOptions.php
```diff
@@ -135,7 +135,7 @@ private function validateRegex($pattern)
         set_error_handler(['Psy\Exception\ErrorException', 'throwException']);
         try {
             preg_match($pattern, '');
-        } catch (ErrorException $e) {
+        } catch (\ErrorException $e) {
```
bobthecow: i think this one is supposed to be `Psy\Exception\ErrorException`, so we're missing a `use` statement at the top, not a slash here.

src/Psy/TabCompletion/Matcher/AbstractContextAwareMatcher.php
```diff
@@ -44,7 +44,7 @@ public function setContext(Context $context)
     /**
      * Get a Context variable by name.
      *
-     * @param $var Variable name
+     * @param Variable $var name
```
bobthecow: this should be `@param string $var Variable name`

test/Psy/Test/TabCompletion/AutoCompleterTest.php
```diff
@@ -21,9 +21,9 @@
 class AutoCompleterTest extends \PHPUnit\Framework\TestCase
 {
     /**
-     * @param $line
-     * @param $mustContain
-     * @param $mustNotContain
+     * @param string $line
+     * @param array  $mustContain
+     * @param array  $mustNotContain
```
bobthecow: see, this is why i don't document tests :P

theofidry: also `array` is not very useful here... Just bump to 7.1 and use typehints 😄

carusogabriel: @bobthecow I don't document they either, but these ones with `dataProvider`s, are important :)

@theofidry Bump to 7.1 is out of the question, I guess :grin:

bobthecow:
Thanks again :)

Issue #416, opened 2017-09-05
Way to purge or reload modified code without exiting?


jchook:
Hey, first.. huge fan of psysh. Thanks for making such a fantastic tool.

My request may already be possible, or may be _im_possible to achieve with PHP. Ultimately it would be really awesome if I could:

1. Start psysh with a config file (to include my autoloader, etc.)
2. Instantiate some classes and do some stuff
3. Encounter an error
4. Go into the code, fix the error, save the file
5. Come back to psysh and re-execute the last command, but using the newly modified code.

Obviously there may need to be a "step 5a." in there where we "purge" the bytecode cache or something. Maybe it's not even possible to do this.

bobthecow:
Unfortunately this isn't really possible. If you're still using 5.x there is a PHP extension called "runkit" which allows this. And PsySH even has support for that! But runkit doesn't support PHP 7, and the closest available option doesn't really support what we'd need for code reloading :(

jchook:
Wow thanks for the tip on runkit.

After Googling, looks like there are some other PHP7 options: [Soft-Mocks](https://github.com/badoo/soft-mocks) and [UPOZ](https://github.com/krakjoe/uopz).

bobthecow:
Unfortunately neither of those allows re-importing files like runkit did.

bobthecow:
Ideally—and with the runkit code reloader this is possible—any existing objects you have would be redefined whenever you update their code.

Something like Soft Mocks works by importing a different version every time you import code, but even if we got it working correctly, it would never update existing objects to match the new code.

So if you create an object, then change its source code, then create an identical object, they won't have the same type. In my opinion, this is a worse place (read: less predictable and less understandable) to be in than having to reload your shell to pick up code changes.

While it doesn't come with a built-in ability to re-import files, UOPZ has at least _some_ of the functionality necessary to do code reloading: If we wrote a fairly sophisticated code reloader, we could delete and redefine methods on existing classes, and change them from private to public or vice versa. We could change the inheritance or interfaces of an already defined class. We could replace functions. And we could redefine constants. All of this would work with objects already instantiated in the shell.

And this might be enough. I've definitely considered exploring it. If you want to take a shot, I'd welcome a pull request :)

jchook:
Man, thanks for the exquisite explanation of all this.

[...]

PR #466, opened 2018-01-24
Feature Proposal: Session Reloading

zachflower:
This PR adds (relatively basic) functionality to reload the current session in order to bring any new file changes without having to exit and re-launch PsySH. This is accomplished by replacing the current process with an identical one using `pcntl_exec`, automatically reloading files and accounting for any changes (see demonstration below).

## Caveats

Because this is a relatively basic implementation of reloading, there are a few caveats that I think are important to mention (and would love any feedback or suggestions to work around them).

### Stateless Reloads

While this reload method will account for any file changes, what it doesn't do is persist the current state, meaning that any methods defined prior to a `reload` will be forgotten.

Because of this, `reload` may not be the most appropriate command name, so if there are any suggestions for something that better reflects what is happening when the command is run I would love to hear them (`refresh` maybe?).

### PCNTL Required

Because this command utilizes `pcntl_exec` to safely replace the current process with a new one, PCNTL _is_ required. While this isn't necessarily a complication (it is a recommend extension, after all), there doesn't seem to be an established convention for commands that rely on optional dependencies.

I would love any thoughts on how to best handle this. As it stands, the current implementation puts a note in the command help text, and throws an exception on execution if `pcntl_exec` isn't available, however disabling the command altogether when it isn't available might be more desirable.

## Demonstration

kapture 2018-01-24 at 14 29 55

## Relevant Issues:

- #416 - Way to purge or reload modified code without exiting?

### DIFF
--- a/src/Command/ReloadCommand.php
+++ b/src/Command/ReloadCommand.php
@@ -28,8 +28,8 @@ class ReloadCommand extends Command
     {
         $this
             ->setName('reload')
-            ->setAliases(array('reload!'))
-            ->setDefinition(array())
+            ->setAliases(['reload!'])
+            ->setDefinition([])
             ->setDescription('Reload the current session.')
             ->setHelp(
                 <<<'HELP'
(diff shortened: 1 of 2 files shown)


zachflower:
Oh snap, definitely mixed feelings! Sad not to have been able to contribute, but delighted that there's something coming down the pike (hopefully something with PHP 7 support coming shortly after) 😄

I will go ahead and close this PR then.

Unrelated: You are one of the most respectful and communicative maintainers I've seen in open source. Just thought I'd say thank you and keep being awesome 🤘

PR #472, opened 2018-03-04
Add ListPass class that validates before cause fatal error

zonuexe:
This CodePass imitates per-version error on list assignment.

## Before

```
% psysh
Psy Shell v0.8.17 (PHP 7.1.14 — cli) by Justin Hileman
>>> [] = []
PHP Fatal error:  Cannot use empty list in /Users/megurine/.composer/vendor/psy/psysh/src/Psy/ExecutionLoop/Loop.php(90) : eval()'d code on line 1

Fatal error: Cannot use empty list in /Users/megurine/.composer/vendor/psy/psysh/src/Psy/ExecutionLoop/Loop.php(90) : eval()'d code on line 1
>>> list() = []
PHP Fatal error:  Cannot use empty list in /Users/megurine/.composer/vendor/psy/psysh/src/Psy/ExecutionLoop/Loop.php(90) : eval()'d code on line 1

Fatal error: Cannot use empty list in /Users/megurine/.composer/vendor/psy/psysh/src/Psy/ExecutionLoop/Loop.php(90) : eval()'d code on line 1
>>> list(1) = [1]
PHP Parse error: Syntax error, unexpected ')' on line 1
>>> list($a) = [1]
=> [
     1,
   ]
```

## After

```
Psy Shell v0.8.17 (PHP 7.1.14 — cli) by Justin Hileman
>>> [] = []
PHP Parse error: Cannot use empty list on line 1
>>> list() = []
PHP Parse error: Cannot use empty list on line 1
>>> list(1) = [1]
PHP Parse error: Syntax error, unexpected ')' on line 1
>>> list($a) = [1]
=> [
     1,
   ]
```

### DIFF
--- a/src/CodeCleaner.php
+++ b/src/CodeCleaner.php
@@ -26,6 +26,7 @@ use Psy\CodeCleaner\ImplicitReturnPass;
 use Psy\CodeCleaner\InstanceOfPass;
 use Psy\CodeCleaner\LeavePsyshAlonePass;
 use Psy\CodeCleaner\LegacyEmptyPass;
+use Psy\CodeCleaner\ListPass;
 use Psy\CodeCleaner\LoopContextPass;
 use Psy\CodeCleaner\MagicConstantsPass;
 use Psy\CodeCleaner\NamespacePass;
@@ -99,6 +100,7 @@ class CodeCleaner
             new InstanceOfPass(),
             new LeavePsyshAlonePass(),
             new LegacyEmptyPass(),
+            new ListPass(),
             new LoopContextPass(),
             new PassableByReferencePass(),
             new ValidConstructorPass(),
--- a/src/CodeCleaner/ListPass.php
+++ b/src/CodeCleaner/ListPass.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ * This file is part of Psy Shell.
+ *
+ * (c) 2012-2018 Justin Hileman
+ *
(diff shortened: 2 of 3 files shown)


bobthecow:
Thank you!

As much as possible, I'd prefer for error messages to match exactly what the uncaught fatal error would have said.

    >>> list("a" => $a) = array("a" => 1)

On PHP 7.0, for example, the actual error from the above code is:

    PHP Parse error:  syntax error, unexpected '=>' (T_DOUBLE_ARROW), expecting :: (T_PAAMAYIM_NEKUDOTAYIM) in php shell code on line 1

But this pass gives us:

    PHP Parse error: Syntax error, unexpected T_CONSTANT_ENCAPSED_STRING, expecting ',' or ')' on line 1

It seems like there might be other cases where this would be true with that same message. Is there an easy(ish) way to get the proper message(s) here?

Also, I couldn't get `Assignments can only happen to writable values` to trigger at all. It looks like you have that error message in the test case, but it's unused. What is it guarding against?

zonuexe:
> Also, I couldn't get `Assignments can only happen to writable values` to trigger at all. It looks like you have that error message in the test case, but it's unused. What is it guarding against?

As you said I overlooked the test for this case.
Following code causes this error in PHP 7.1+: https://3v4l.org/rNFB0

```php
<?php ["a"] = [1];
```

bobthecow:
Thanks again for this. I've resolved the merge conflicts and merged it into `develop` :)

PR #475, opened 2018-03-25
Avoid conflicts with PHP Parser v4

m1guelpf:


bobthecow:
Please make this change against the `develop` branch :)

m1guelpf:
@bobthecow Done!

bobthecow:
Unfortunately it's not _quite_ this simple :( 

[PHP Parser 4.x has a bunch of backwards compatibility breaks](https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-4.0.md). I've got fixes for the ones exposed by the PsySH test suite, but at runtime we need to accept all manner of craziness from users, so I'd like to spend some time going through that list and making sure we won't be caught by things that aren't covered.

See #476

PR #476, opened 2018-03-25
Add support for PHP Parser 4.x.

bobthecow:
There are a decent number of backwards compatibility breaks in 4.0:

https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-4.0.md

This handles everything exposed by the PsySH test suite, but at runtime we need to accept all manner of craziness from users, so I'd like to spend some time going through that list and making sure we won't be caught by things that aren't covered.

Fixes #475


GrahamCampbell:
Is there anything we can do to easily increase the test coverage? I noticed on the previous PR, no tests failed. :P

bobthecow:
Something was wrong with the previous PR. The code fixed here _definitely_ failed before I opened the PR :)

bobthecow:
For example… 

https://travis-ci.org/bobthecow/psysh/jobs/358195657

bobthecow:
Yes, more code coverage would be great. The code affected here has about the best coverage in the whole project, though. If you ignore branches that aren't relevant in PHP 7.x and PHP Parser 4.x, the CodeCleaner code has 98%+ coverage :)

przemoli:
Is this code slated for the next release? What features are potentially affected by this change?
(New to Psysch, so I can't test for BC, but I will explore on my own anyway, so I can do some testing too)

bobthecow:
this is slated for the next release. it was merged into the development branch last week, which is where stable releases are cut from :)

PR #480, opened 2018-04-05
Update composer.json

mercuryseries:
Allow also ~4.0


bobthecow:
It’s not quite that simple, since there are backwards compatibility breaks in 4.0 (see #476). 

But! They’re already taken care of. You can use the `develop` branch (or `0.9.x-dev` via Composer) if you need it right now.

PR #481, opened 2018-04-15
Restore symfony console 2.3 support

GrahamCampbell:
Was there a reason this was removed?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/console": "^2.4.2|~3.0|~4.0",
+        "symfony/console": "^~2.3.10|2.4.2|~3.0|~4.0",
         "symfony/var-dumper": "~2.7|~3.0|~4.0",
         "nikic/php-parser": "~2.0|~3.0|~4.0",
         "dnoegel/php-xdg-base-dir": "0.1",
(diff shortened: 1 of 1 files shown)


bobthecow:
I'm not sure, I probably just missed it when merging the PHP 5.4+ PR.

Tested and it works great, so let's add it back!

GrahamCampbell:
👍

bobthecow:
In keeping with our long, rich tradition of releasing `v0.x.1` as fast as possible after `v0.x.0`, this has gone out as `v0.9.1`. Thanks for catching it 😄

GrahamCampbell:
🚢

GrahamCampbell:
So is php 5.4, and 5.5, and symfony 2.4, 2.5, and 2.6 was ages ago. Why does that matter? It's not the job of this package to force people not to use them. If it works fine, there's no real case for not supporting them.

GrahamCampbell:
> Seriously, update if you can.

If people can, then they will. Nobody's saying people have to use 2.3, but if they need too, then can. ;)

GrahamCampbell:
btw https://symfony.com/roadmap?version=2.4#checker :P

Taluu:
Didn't even mention 2.4, 2.5 and 2.6 because 2.3 was a lts that was finished a year ago. Yes, it is indeed a library to not push people to upgrade. But it's not a library's job either to support unmaintained libraries.

PR #486, opened 2018-04-23
make doc command support complete

lvht:


lvht:
Hi @bobthecow , this patch does not work for `doc Composer\Autoload\ClassLoader::<tab>`. Could you offer any advice?

bobthecow:
Why should `doc` be special? What about `ls` and `show`? :)

lvht:
@bobthecow support ls and show

lvht:
ping @bobthecow

bobthecow:
Sorry about the delay. I finally tracked down and fixed the namespaced class property and method completion! (It was an existing bug; see 1f9bc08)

This looks good to me.

bobthecow:
Thanks for your contribution. This will go out in the next stable release.

PR #492, opened 2018-05-07
Develop is now 0.10

GrahamCampbell:
I think?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
     "bin": ["bin/psysh"],
     "extra": {
         "branch-alias": {
-            "dev-develop": "0.9.x-dev"
+            "dev-develop": "0.10.x-dev"
         }
     }
 }


bobthecow:
There aren't any backwards compatibility breaks in `develop` so far, so if I were to cut a release right now it'd be 0.9.4.

bobthecow:
Yeah, the isolated version would likely be v0.10.

GrahamCampbell:
> There aren't any backwards compatibility breaks in develop so far, so if I were to cut a release right now it'd be 0.9.4.

Doesn't semver dictate new features require a minor version bump?

bobthecow:
>Doesn't semver dictate new features require a minor version bump?

Not for v0.x.

bobthecow:
The real answer is that I should release a v1.0 some day and then Server will do what you expect ;)

Closing for now.

PR #493, opened 2018-05-08
Makefile improvements

bobthecow:
* Extract composer opts into variables.
* Use composer to set required PHP version for phar builds.
* Minor help output cleanup.
* Use Box output config to save a cp / mv step in phar builds.
* Only delete build folder contents when cleaning (so it doesn't do dumb things when i've got a terminal open to that directory)
* Replace bin/package with make
* Always compile the phar (save ourselves from failures only when it comes time to tag)
* Switch to make for travis deploys... and only run on the latest PHP since we build all of the versions in one shot.


bobthecow:
@theofidry ☝️ 😄

Makefile
```diff
@@ -46,46 +57,59 @@ build/psysh: bin/psysh src composer.json composer.lock box.json.dist build/stub
 	rm -rf build/psysh || true
 	mkdir build/psysh
 	cp -R $(PSYSH_SRC) build/psysh/
-	sed -i '' 's/"php": ".*"/"php": ">=7.0.0"/g' build/psysh/composer.json
 	composer config --working-dir build/psysh platform.php 7.0
-	composer update --working-dir build/psysh --prefer-stable --no-dev --no-progress --classmap-authoritative --no-interaction --verbose --prefer-dist
+	composer require --working-dir build/psysh $(COMPOSER_REQUIRE_OPTS) php:'>=7.0.0'
```
theofidry: indeed that's more elegant...

bobthecow:
Thanks!

PR #497, opened 2018-05-16
A bunch more Makefile cleanup.

bobthecow:
See individual commits for more commentary :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,25 +9,23 @@ COMPOSER_REQUIRE_OPTS = $(COMPOSER_OPTS) --no-update
 COMPOSER_UPDATE_OPTS = $(COMPOSER_OPTS) --prefer-stable --no-dev --classmap-authoritative --prefer-dist
 
 help:
-	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/## *//'
+	@echo "\033[33mUsage:\033[0m\n  make TARGET\n\n\033[33mTargets:\033[0m"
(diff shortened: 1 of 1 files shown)


bobthecow:
cc @theofidry

Makefile
```diff
@@ -1,115 +1,96 @@
 .DEFAULT_GOAL := help
+.PHONY: help clean build dist
```
theofidry: the reason I went with multiple PHONY is that it's easy to check you didn't forget one. You see a rule and you can see if it's a command by the presence of the PHONY target. Without it you easily add a new command and completely forget about it

Makefile
```diff
@@ -1,115 +1,96 @@
...
-# Rules from files
-#---------------------------------------------------------------------------
+# All the composer stuffs
 
 composer.lock: composer.json
-	@echo The composer.lock file is not synchronized with the composer.json file
-
-vendor: composer.lock
 	composer install
+	touch $@
```
theofidry: nice trick

Makefile
bobthecow: Yeah, I can see that. But you could also add a new command down at the bottom of the file, and never thing about adding .PHONY because none of the targets down there have it :) 

I feel like having cleaner command definitions outweighs downsides.

I did just move the .PHONY and .DEFAULT_GOAL next to the list of commands, since that's what they're used for. That is probably a good balance?

theofidry: I think it's ok either way. You don't have that many commands so it's ok, I've been bitten a couple of times by some missing `.PHONY` target and easier to manage the other way but I also had more commands, so harder to track

PR #509, opened 2018-07-01
Add a polyfill for the ctype extension

BackEndTea:
A cytpe function is used here: https://github.com/bobthecow/psysh/blob/b8ea8820cd6ea4ea60a40b86ca16aa4c1a237152/src/Util/Docblock.php#L225

But ctype is not always enabled, especially for minimalist distributions like alpine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "php": ">=5.4.0",
         "symfony/console": "~2.3.10|^2.4.2|~3.0|~4.0",
         "symfony/var-dumper": "~2.7|~3.0|~4.0",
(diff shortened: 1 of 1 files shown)


bobthecow:
Heh. Thanks!

I figured basically everything had the ctype extension so I didn't worry about including it as a dependency. If there are distros that don't have it, it's simple enough to use a regular expression for that :) 

I've removed the one call to `ctype_alpha` in 4b48964bbbde4165666d9eb632fc3783e54f24fa and that'll go out in the next release.

I've also checked against all the other polyfills Symfony provides, and core PsySH code doesn't have a dependency on any of them. So thanks for the nudge.

PR #517, opened 2018-08-10
Display CWD of root script calling Psy Shell in header

brysonreece:


brysonreece:
Not sure why StyleCI checks are failing. Looks like a whitespace issue, but the whitespace is formatted correctly. (Maybe a tabs v. spaces issue?)

bobthecow:
I think most people likely don't have multiple versions of psysh installed, and most of the time this will say `cli in bin`, so it's probably not super useful information in general, but you can totally add this to _your_ welcome message, by adding this to your psysh config:

```php
<?php

return [
  'startupMessage' => sprintf(
    '<info>Running in %s</info>', basename(dirname($_SERVER['PHP_SELF']))
  ),
];
```

… or something like that :)

PR #518, opened 2018-08-12
Accept static method call of trait

zonuexe:
Trait's static method can be called without class mixined.

```php
<?php

trait Foo
{
    public static function bar()
    {
        return __METHOD__;
    }
}

var_dump(Foo::bar());
```

https://3v4l.org/hKUHe


bobthecow:
Thanks!

PR #527, opened 2018-10-08
Bump Box

theofidry:
- Bump box to the latest stable (3.1.0)
- Remove an unnecessary setting (the value was the default value)

### DIFF
--- a/box.json.dist
+++ b/box.json.dist
@@ -4,7 +4,6 @@
     "compactors": [
         "KevinGH\\Box\\Compactor\\Php"
     ],
-    "chmod": "0755",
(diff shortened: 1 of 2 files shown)


theofidry:
@bobthecow is the Codecov report in PRs really necessary? I think the link in the checks should be enough, no need for a message

bobthecow:
You're right. That codecov report is kind of obnoxious. But I can't figure out if there's a good way to disable it :( 

Do you mind pointing this PR at `develop` rather than `master`? I think that'll fix the negative coverage, too.

theofidry:
Done :)

PR #534, opened 2018-11-11
Some PHPStan fixes

carusogabriel:
Following #460


src/Command/Command.php
```diff
@@ -243,7 +243,7 @@ private function formatDefaultValue($default)
      *
      * Falls back to legacy TableHelper.
      *
-     * @return Table|TableHelper
+     * @return TableHelper
```
bobthecow: This one is incorrect; the method returns either `Table` or `TableHelper`.

bobthecow:
Thanks!

PR #535, opened 2018-11-13
Exclude files and folder from release archive

ankurk91:
Hi,

Let's not publish all those file.
`.gitattributes` can be used to exclude files from final archive that is downloaded by composer.

Read more
- https://madewithlove.be/gitattributes/
- https://alexbilbie.com/2012/11/exclude-objects-with-gitattributes/
- http://www.pixelite.co.nz/article/using-git-attributes-exclude-files-your-release/


reedy:
Hey. How come this never made it into master? Any chance we can cherry pick the commit in? :)

bobthecow:
Because there wasn't a stable release yet 😬 

🔜‼️

PR #537, opened 2018-11-21
Move PHPUnit to a composer bin dependency

bobthecow:
We aren't super picky about which version of PHPUnit we use, and really there is no reason to complicate our runtime dependencies by taking on all of PHPUnit's dependencies as well.

So now we can use globally-installed PHPUnit, or a local phar, or a composer-bin-installed version. Basically anything but a dev dependency :)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -7,3 +7,7 @@ PsySH follows [PSR-1](http://php-fig.org/psr/psr-1/) and [PSR-2](http://php-fig.
 ## Branching model
 
 Please branch off and send pull requests to the `develop` branch.
+
+## Running tests
+
+1. Run `make test` to run tests!
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,43 +3,45 @@ language: php
 sudo: false
 
 matrix:
-  include:
-    - php: 5.4
-    - php: 5.4
-      env: 'COMPOSER_FLAGS="--prefer-lowest --prefer-stable"'
-    - php: 5.5
-    - php: 5.6
-    - php: 7.0
-    - php: 7.1
-    - php: 7.2
-    - php: 7.3
-    - php: hhvm
-      dist: trusty
-  allow_failures:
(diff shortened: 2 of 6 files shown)


bobthecow:
@theofidry @GrahamCampbell any reason not to do this?

.github/CONTRIBUTING.md
```diff
@@ -7,3 +7,8 @@ PsySH follows [PSR-1](http://php-fig.org/psr/psr-1/) and [PSR-2](http://php-fig.
 ## Branching model
 
 Please branch off and send pull requests to the `develop` branch.
+
+## Running tests
+
+1. Run `composer bin phpunit install` to install PHPUnit.
```
theofidry: how about moving this into the `Makefile`?

```
PHPUNIT=vendor/bin/phpunit

.PHONY test
test: $(PHPUNIT)

$(PHPUNIT)t: vendor-bin/phpunit
	composer bin phpunit install
	touch $@
```

This way the DX is not affected by it either

.travis.yml
```diff
@@ -24,7 +24,7 @@ matrix:
 install: travis_retry composer update --no-interaction $COMPOSER_FLAGS
 
 script:
- - vendor/bin/phpunit --verbose --coverage-clover=coverage.xml
+ - phpunit --verbose --coverage-clover=coverage.xml
```
theofidry: I think you are currently missing the install step for PHPUnit, but moving it to the makefile would also solve that issue IMO.

bobthecow: We're using the system-installed PHPUnit on CI. I'm pretty sure there's no one version of PHPUnit that can work on all the versions of PHP we currently test, so using composer bin there is a bit problematic.

Unfortunately, the Travis system-installed PHPUnit is outdated on PHP 5.5, apparently, so the build is failing.

.github/CONTRIBUTING.md
bobthecow: Yeah, makes sense.

PR #538, opened 2018-11-21
shell: use default foreground on error, etc

apreiml:
I suggest to use the default foreground colors on error, waring and hidden, otherwise those messages are harder to read on dark terminal themes.

Example: Default on the left side, black on the right side.

2018-11-21-155011_slurp_grim

### DIFF
--- a/src/Output/ShellOutput.php
+++ b/src/Output/ShellOutput.php
@@ -174,13 +174,13 @@ class ShellOutput extends ConsoleOutput
     {
         $formatter = $this->getFormatter();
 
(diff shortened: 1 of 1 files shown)


bobthecow:
But it also can make things _harder_ to read.

Screen Shot 2019-10-03 at 10 16 38 PM

TBH I'd consider this a bug in your color palette. Ideally, the standard foreground and background colors can be used in any combination, including inverting them like PsySH does for errors and warnings.

Color palettes like [Solarized](https://ethanschoonover.com/solarized/) and [base16](http://chriskempson.com/projects/base16/) are built around this premise :)

apreiml:
Ah ok. Thanks for looking into this!

PR #553, opened 2019-03-08
Set readline_name

tgr:
This allows application-specific bindings in `.intputrc` or `.editrc`.

In `.editrc`, you can set psysh-specific bindings with a `psysh:` prefix:

    psysh:bind "\e[5~" ed-search-prev-history

In `.inputrc`, you can use [conditional init constructs](https://www.gnu.org/software/bash/manual/html_node/Conditional-Init-Constructs.html) (in theory; I don't have readline set up so I didn't test this):

    $if psysh
    "\e[5~": history-search-backward
    $endif

Unconditional/unprefixed bindings will still work.


bobthecow:
Thank you! This has been merged into the `develop` branch and will go into the next release :)

PR #560, opened 2019-05-24
Update default editor to use vi instead of nano

jycr753:
- The reason is because `vi` is commonly install by default in light weight Linux distribtuions


bobthecow:
unfortunately there isn't a great universal answer. nano is installed by default on Ubuntu-based linux flavors (and vim isn't). 

really what it comes down to is that nobody has a hard time saving or exiting in nano, and vim is famously unfriendly to people who accidentally end up in it. given that, i'm not comfortable with changing the default.

but if you want to use vim, please set your `$EDITOR` environment variable up properly and psysh will respect it :)

PR #566, opened 2019-08-20
Promoting good user experience

talbergs:
Author might decide better place, where to put this information. Try it out!


bobthecow:
This doesn't really belong in the README, as that's just a table of contents for the wiki. Maybe there's somewhere in the wiki we can put this sort of thing?

PR #577, opened 2019-11-23
Sf5 compatibility

Taluu:
Add compatibility for Symfony 5.

I had to "fix" a call on a few `write` call from the Output to match the signature that is used since ~2.0 of symfony.

#SymfonyHackday

### DIFF
--- a/src/Command/HistoryCommand.php
+++ b/src/Command/HistoryCommand.php
@@ -140,11 +140,13 @@ HELP
         } else {
             $type = $input->getOption('no-numbers') ? 0 : ShellOutput::NUMBER_LINES;
             if (!$highlighted) {
-                $type = $type | ShellOutput::OUTPUT_RAW;
+                $type = $type | OutputInterface::OUTPUT_RAW;
             }
 
             $output->page($highlighted ?: $history, $type);
         }
+
+        return 0;
     }
 
     /**
--- a/src/Command/ShowCommand.php
+++ b/src/Command/ShowCommand.php
@@ -18,7 +18,6 @@ use Psy\Exception\RuntimeException;
 use Psy\Formatter\CodeFormatter;
 use Psy\Formatter\SignatureFormatter;
 use Psy\Input\CodeArgument;
-use Psy\Output\ShellOutput;
 use Symfony\Component\Console\Formatter\OutputFormatter;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -99,11 +98,15 @@ HELP
                 throw new \InvalidArgumentException('Too many arguments (supply either "target" or "--ex")');
             }
 
-            return $this->writeExceptionContext($input, $output);
+            $this->writeExceptionContext($input, $output);
+
+            return 0;
         }
 
         if ($input->getArgument('target')) {
-            return $this->writeCodeContext($input, $output);
+            $this->writeCodeContext($input, $output);
+
+            return 0;
         }
 
         throw new RuntimeException('Not enough arguments (missing: "target")');
@@ -117,7 +120,7 @@ HELP
         $this->setCommandScopeVariables($reflector);
 
         try {
-            $output->page(CodeFormatter::format($reflector, $this->colorMode), ShellOutput::OUTPUT_RAW);
+            $output->page(CodeFormatter::format($reflector, $this->colorMode), OutputInterface::OUTPUT_RAW);
         } catch (RuntimeException $e) {
             $output->writeln(SignatureFormatter::format($reflector));
             throw $e;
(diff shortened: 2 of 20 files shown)


Taluu:
Failures unrelated, as travis doesn't support 5.4 and 5.5 anymore (neither should you, but I know your stance on that, and I think you know mine. :D)

GrahamCampbell:
Travis does support those versions. You jut need to tell it to use an older dist.

Example: https://github.com/schmittjoh/php-option/blob/1.5.2/.travis.yml.

GrahamCampbell:
Please target the develop branch of this repo for your PR instead.

Taluu:
Fixed.

GrahamCampbell:
You;'ve broken support for Symfony 3 now though?

GrahamCampbell:
And some versions of 4.

src/Command/HistoryCommand.php
```diff
@@ -140,11 +140,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
         } else {
             $type = $input->getOption('no-numbers') ? 0 : ShellOutput::NUMBER_LINES;
             if (!$highlighted) {
-                $type = $type | ShellOutput::OUTPUT_RAW;
+                $type = $type | OutputInterface::OUTPUT_RAW;
```
GrahamCampbell: Will this work on Symfony 3?

src/Command/ShowCommand.php
```diff
@@ -216,7 +219,7 @@ private function writeTraceCodeSnippet(OutputInterface $output, array $trace, $i
             return;
         }
 
-        $output->write($this->getHighlighter()->getCodeSnippet($code, $line, 5, 5), ShellOutput::OUTPUT_RAW);
+        $output->write($this->getHighlighter()->getCodeSnippet($code, $line, 5, 5), false, OutputInterface::OUTPUT_RAW);
```
GrahamCampbell: This won't work on Symfony 3.

src/Command/WhereamiCommand.php
```diff
@@ -126,11 +125,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $output->writeln('');
         $output->writeln(\sprintf('From <info>%s:%s</info>:', $this->replaceCwd($info['file']), $info['line']));
         $output->writeln('');
-        $output->write($highlighter->getCodeSnippet($contents, $info['line'], $num, $num), ShellOutput::OUTPUT_RAW);
+        $output->write($highlighter->getCodeSnippet($contents, $info['line'], $num, $num), false, OutputInterface::OUTPUT_RAW);
```
GrahamCampbell: Same

src/Command/ShowCommand.php
Taluu: It will : the argument is present since the very first commit of the interface, which was introduced in august 2010.

Taluu: (So this means this was faulty to begin with...)

src/Command/WhereamiCommand.php
Taluu: See my previous remark.

src/Command/HistoryCommand.php
Taluu: Yes, the OutputInterface is present since at least 2.0 : https://github.com/symfony/console/commit/44124245351cbf722f051a3d4f04a9640ebb07a3

src/Command/ShowCommand.php
Taluu: See https://github.com/symfony/console/commit/68c191cd02f0dbffeef342e95c57cf741008abfd

src/Command/Command.php
```diff
@@ -283,7 +283,7 @@ protected function getTableHelper()
         return $table
             ->setRows([])
             ->setLayout(TableHelper::LAYOUT_BORDERLESS)
-            ->setHorizontalBorderChar('')
+            ->setHorizontalBorderChars('')
```
bobthecow: `getTableHelper` is only called for ~2.4 TableHelpers, which uses `setHorizontalBorderChar`:

https://github.com/symfony/console/blob/2.4/Helper/TableHelper.php#L224

src/Command/ShowCommand.php
Taluu: Found the real commit which introduced this parameter : https://github.com/symfony/symfony/commit/e715bc3d45496e77d52b497ef65ef98aa1e99906

One question pops to mind though : will we want to add a new line at the end or not ? As if the `ShellOutput::OUTPUT_RAW` (which is `OutputInterface::OUTPUT_RAW` : it is imo less misleading to use the proper class for the constants, especially if they aren't overwritten - I'm not even sure it's actually possible) was `1`, then maybe I should use `true` instead to avoid a BC break ?

WDYT @bobthecow, @GrahamCampbell ?

src/Command/Command.php
Taluu: Should we maybe then "deprecate" this helper call ? 2.4 (even 2.x) is not maintained anymore...

Taluu: (Left it as is then)

src/Command/ShowCommand.php
bobthecow: In this particular case, it doesn’t make a difference. The output is the same either way. I’d probably lean toward `false`.

bobthecow:
Thanks @Taluu!

As mentioned in https://github.com/laravel/tinker/pull/80 I’m going to probably cut a v0.9.x release with this as well.

PR #579, opened 2019-11-23
add Symfony 5.0 support #SymfonyHackday

mikvet:
Hi there, 
During the #SymfonyHackday I updated your composer.json to support Symfony 5 which has been released. 
Tests are passing :-)


mikvet:
better PR already submited, have fun discussing versions ;-)

PR #580, opened 2019-11-27
Add missing ShellOutput import

funkjedi:
Removal of import in 0.9.10 broken things as the class is still being used here:

https://github.com/bobthecow/psysh/blob/5442009e4ac8ad01ed7e06823350bd75e692257a/src/Shell.php#L374


bobthecow:
Hey, thanks for noticing! I pulled the tag / release as soon as I saw it, and everything is fixed in v0.9.11 😬

PR #582, opened 2019-12-01
Test on HHVM 3 insead of 4

GrahamCampbell:
HHVM has intentionally stopping being compatible with PHP as of v4.0. Composer won't even run on HHVM 4. We should test on the last LTS release of HHVM 3, which is the version anyone using HHVM who uses lots of PHP libs (and not hacklang) will be using.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
         - php: 7.2
         - php: 7.3
         - php: 7.4
-        - php: hhvm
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
We probably need to force PHPUnit 4 on HHVM,

bobthecow:
Honestly, if PHPUnit doesn't even work in it, I'm not sure how much use it is to keep testing :-/

PR #583, opened 2019-12-02
Fixed Travis Config

GrahamCampbell:
Fixes PHP 7.4 and HHVM test running.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,18 @@ matrix:
         - php: 5.5
           dist: trusty
         - php: 5.6
+          dist: xenial
         - php: 7.0
+          dist: xenial
         - php: 7.1
+          dist: bionic
         - php: 7.2
+          dist: bionic
         - php: 7.3
-        - php: 7.4snapshot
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,12 +12,18 @@ matrix:
...
+          dist: bionic
         - php: 7.2
+          dist: bionic
         - php: 7.3
-        - php: 7.4snapshot
-        - php: hhvm
+          dist: bionic
+        - php: 7.4
+          dist: bionic
+        - php: hhvm-3.18
```
bobthecow: let's update the `allowed_failures` entry to keep this from making too much noise?

GrahamCampbell:
> any downside to explicitly specifying a dist for everything?

Only can be an advantage, so that when travis changes the default dist in the future, our config doesn't break. For example, they recently changed it from trusty to xenial, which broke older PHP 5 builds.

.travis.yml
GrahamCampbell: Oh yes.

PR #585, opened 2019-12-04
Fix unreadable error message

kristianklok:
The error messages turn out as grey on red on some systems making them unreadable without highlighting. Suggest switching to white on red.


GrahamCampbell:
I think this PR needs to be rebased again. ;P

GrahamCampbell:
Against the develop branch, rather than master.

PR #587, opened 2019-12-05
Fix unreadable error messages on some systems

kristianklok:
The error message has black on red text which turns unreadable on some systems, Suggest changing it to white omn red.


shinyhiram:
Just leaving a screenshot of what it looks like on an affected system:
Screen Shot 2019-12-17 at 4 04 42 PM

PR #589, opened 2019-12-05
Fix PHP 7.4 deprecation: Unparenthesized `a ? b : c ? d : e` is deprecated

Salmatron:


bobthecow:
Hey, thanks for the PR! This is fixed in the development branch, and will be in the next stable release. In the meantime, I'd suggest using a dev dependency target.

Issue #581, opened 2019-12-01
PHP-Scoper incorrectly scopes `\ReflectionClassConstant`


bobthecow:
`ReflectionClassConstant::create` tries to decide whether to create an instance of the actual `\ReflectionClassConstant`, or our ponyfill. When trying to re-enable `PHP-Scoper` it was incorrectly scoping that class name, which caused test failures:

```php
    public static function create($class, $name)
    {
        if (\class_exists('_HumbugBox4f95cd3d5edb\\ReflectionClassConstant')) {
            return new \ReflectionClassConstant($class, $name);
        }
        return new self($class, $name);
    }
```

cc @theofidry

theofidry:
Maybe it should be fixed via a patched; Also maybe PhpScoper should be
enabled for the most cutting edge PHAR and omit it for the others where you
include various polyfills

bobthecow:
In this case, the ReflectionClassConstant bit is part of PsySH core and is in everything, not optionally just in compatibility builds.

This feels like it might be a bug in PHP Scoper, tbh. the original code was

```php
if (\class_exists('\\ReflectionClassConstant')) {
  // ...
```

… which couldn't be anything but a reference to an internal class.

theofidry:
Hm indeed

bobthecow:
@theofidry is there an easy way to work around this on our end while we're waiting for the issue in PHP Scoper to be resolved?

theofidry:
Yes, by relying on a [patcher](https://github.com/humbug/php-scoper#patchers)

PR #592, opened 2019-12-13
Fix `ReflectionClassConstant` class_exist() call

theofidry:
Closes https://github.com/bobthecow/psysh/issues/581

I'm fixing the issue on PHP-Scoper side as well, but the idea is that in `class_exists()` and similar functions, symbol paths are always the fully qualified names so the leading `\` is unnecessary

### DIFF
--- a/src/Reflection/ReflectionClassConstant.php
+++ b/src/Reflection/ReflectionClassConstant.php
@@ -219,7 +219,7 @@ class ReflectionClassConstant implements \Reflector
      */
     public static function create($class, $name)
     {
-        if (\class_exists('ReflectionClassConstant')) {
+        if (\class_exists(\ReflectionClassConstant::class)) {
             return new \ReflectionClassConstant($class, $name);
         }
 
--- a/src/Reflection/ReflectionClassConstant.php
+++ b/src/Reflection/ReflectionClassConstant.php
@@ -219,7 +219,7 @@ class ReflectionClassConstant implements \Reflector
      */
     public static function create($class, $name)
     {
-        if (\class_exists(\ReflectionClassConstant::class)) {
+        if (\class_exists('ReflectionClassConstant')) {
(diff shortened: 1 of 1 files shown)


src/Reflection/ReflectionClassConstant.php
```diff
@@ -219,7 +219,7 @@ public function getEndLine()
      */
     public static function create($class, $name)
     {
-        if (\class_exists('\\ReflectionClassConstant')) {
+        if (\class_exists('ReflectionClassConstant')) {
```
GrahamCampbell: Should be `\ReflectionClassConstant::class`

src/Reflection/ReflectionClassConstant.php
```diff
@@ -219,7 +219,7 @@ public function getEndLine()
      */
     public static function create($class, $name)
     {
-        if (\class_exists('\\ReflectionClassConstant')) {
+        if (\class_exists(\ReflectionClassConstant::class)) {
```
bobthecow: `::class` doesn’t work on 5.4.x which is still our minimum version :)

Also it’s super weird to me that PHP lets you do this. If the class doesn’t exist, which is what you’re checking for, _where did that class constant come from_.

GrahamCampbell: Oh, I thought the min version was 5.5 for some reason. Sorry!

GrahamCampbell: Maybe it's time to drop PHP 5.4 and 5.5 on `develop` now, anyway?

bobthecow:
Thanks!

Does this mean we can remove our scoper hack now, or do we have to wait for a change in scoper?

theofidry:
This path should remove the need of the patcher workaround with the current
version of PHP-Scoper and Box. But I’ve released both as well so even
without my patch, just updating them will fix the problem

PR #594, opened 2019-12-14
 Drop support for PHP 5.4 and HHVM less than 3.18 

GrahamCampbell:
1. Raise minimum PHP version to 5.5.9;
2. Raise minimum HHVM version to 3.18;
3. Remove legacy code to handle versions older than above;
4. Updated builds so compat builds with PHP 5.5.9 and main version with 7.2.5;
5. Updated bin script to ~check the version and~ write errors to stderr instead of stdout.
6. EDIT: Check runtime version and extensions in the bin function.
7. EDIT: Test on PHP 8 too.

Possible future changes (but maybe right now is still too soon):
1. Drop Symfony 2;
2. Drop PHPParser 1.
3. Drop HHVM entirely.
4. Drop PHP 5.5 too.


src/CodeCleaner/FunctionReturnInWriteContextPass.php
```diff
@@ -64,11 +55,8 @@ public function enterNode(Node $node)
                     continue;
                 }
 
-                $msg = ($node instanceof Isset_ && $this->atLeastPhp55) ? self::PHP55_MESSAGE : self::EXCEPTION_MESSAGE;
-                throw new FatalErrorException($msg, 0, E_ERROR, null, $node->getLine());
+                throw new FatalErrorException(self::ISSET_MESSAGE, 0, E_ERROR, null, $node->getLine());
```
bobthecow: this still needs to distinguish between the two messages (based on `$node instanceof Isset_`).

we could prolly rename the constants to something clearer though :)

bin/psysh
```diff
@@ -89,27 +98,16 @@ if (!class_exists('Psy\Shell')) {
...
 }
 
 // If the psysh binary was included directly, assume they just wanted an
 // autoloader and bail early.
-//
-// Keep this PHP 5.3 code around for a while in case someone is using a globally
-// installed psysh as a bin launcher for older local versions.
-if (version_compare(PHP_VERSION, '5.3.6', '<')) {
-    $trace = debug_backtrace();
-} elseif (version_compare(PHP_VERSION, '5.4.0', '<')) {
```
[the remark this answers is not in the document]
GrahamCampbell: The reason for deleting this is because the earlier part of the script ensures the correct version of PHP is running.

src/CodeCleaner/FunctionReturnInWriteContextPass.php
GrahamCampbell: Yeh, I messed this up, then didn't have time to fix it earlier today. I'll do it soon.

GrahamCampbell: Fixed, and updated the message to match the message PHP 7.0.0 and higher will spit out.

src/functions.php
```diff
@@ -272,6 +272,38 @@ function info(Configuration $config = null)
     function bin()
     {
         return function () {
+            if (!isset($_SERVER['PSYSH_IGNORE_ENV']) || !$_SERVER['PSYSH_IGNORE_ENV']) {
```
bobthecow: This (and the helpful error messages) is a great call.

GrahamCampbell: Borrowed the idea from php-cs-fixer, actually. ;)

GrahamCampbell: The idea of the env variable that is, not the error messages. :P

GrahamCampbell:
Ok, let's see what travis makes of my most recent change. Since our bin script says PHP 8 is good to go, we should test it. ;)

bobthecow:
Maybe nightly should be an allowed failure even if it passes now? I guess we could change it if/when it gets flaky.

GrahamCampbell:
I think maybe see how things go, and if things start to fail, we can either fix them, or allow failures. If we allow failures from the start, we might not notice for a while when things do start to fail, if they do. The CI should help us notice fast. ;)

Makefile
```diff
@@ -51,7 +51,7 @@ vendor/bin/box: vendor/autoload.php
 	touch $@
 
 vendor/bin/phpunit: vendor/autoload.php
-	composer bin phpunit install
+	composer bin phpunit install --ignore-platform-reqs
```
GrahamCampbell: This change is safe because of the lock file.

GrahamCampbell:
Hmmm, I wonder if https://github.com/bobthecow/psysh/commit/1ce74a427e81dbc68e98f66a3fe446e9aa913621 also fails on the latest PHP 7.4.x snapshot. The change doesn't seem needed on PHP 8.

GrahamCampbell:
Yeh, looks like the change was actually reverted in PHP 7.4.1. Travis is still running PHP 7.4.0.

GrahamCampbell:
All green now. ;)

bobthecow:
Thank you :)

PR #605, opened 2020-02-07
Support PHP7.4

xepozz:
## How to roll?

1. Temporary add showing all errors `error_reporting(E_ALL);` to `SignatureFormatterTest` class.
2. Run test with flag `--fail-on-warning`, like that: `php vendor/bin/phpunit --fail-on-warning`

## Fixed
https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.nested-ternary

## Additional impovements
Added `.idea` folder to `.gitignore`
Added PHP 7.3 and PHP 7.4 to Travis CI building.
Added `bootstrap` for tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 /build/
 /dist/
 /composer.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ matrix:
     - php: 7.0
     - php: 7.1
     - php: 7.2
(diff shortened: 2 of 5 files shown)


GrahamCampbell:
This is already fixed on develop. ;)

xepozz:
Ok, when it will be realesed?

GrahamCampbell:
Not sure. You can install it using `^0.10`.

PR #612, opened 2020-03-22
Removed extra method

GrahamCampbell:
`_redisplay` is not in the interface, and has a non-PSR name. Why was it added?

### DIFF
--- a/src/Readline/HoaConsole.php
+++ b/src/Readline/HoaConsole.php
@@ -37,7 +37,11 @@ class HoaConsole implements Readline
     public function __construct()
     {
         $this->hoaReadline = new HoaReadline();
-        $this->hoaReadline->addMapping('\C-l', [$this, '_redisplay']);
+        $this->hoaReadline->addMapping('\C-l', function () {
+            $this->redisplay();
+
+            return HoaReadline::STATE_NO_ECHO;
+        });
     }
 
     /**
--- a/src/Readline/HoaConsole.php
+++ b/src/Readline/HoaConsole.php
@@ -101,19 +101,9 @@ class HoaConsole implements Readline
      */
     public function redisplay()
     {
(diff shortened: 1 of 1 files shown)


src/Readline/HoaConsole.php
```diff
@@ -100,20 +100,10 @@ public function readline($prompt = null)
...
-    /**
-     * @return int
-     */
-    public function _redisplay()
     {
         $currentLine = $this->hoaReadline->getLine();
         Cursor::clear('all');
         echo $this->lastPrompt, $currentLine;
-
-        return HoaReadline::STATE_NO_ECHO;
```
zonuexe: See ba7c3baef1dcb403716912a62bb26dddffe3f50a 342093a8aa3a9e8cad90053e6d88751f1fa6d8f3

Do not remove this return value.  If this is omitted, a line feed will be output.

GrahamCampbell: You don't return that value anyway in the original code?

GrahamCampbell:
```
    public function redisplay()
    {
        $this->_redisplay();
    }
```

zonuexe:
These methods have been split to keep [`Readline::redisplay()`](https://github.com/bobthecow/psysh/blob/v0.10.2/src/Readline/Readline.php#L68) interface. This interface does not have `@return`, but I believe it should be interpreted as `@return void` rather than`@return mixed`.

src/Readline/HoaConsole.php
GrahamCampbell: Ah, I see the call. Fixed.

GrahamCampbell:
I've made a fix now. We really shouldn't have this extra public method as a hack when we don't need it.

zonuexe:
LGTM

PR #615, opened 2020-03-22
Fixed configuration tests

GrahamCampbell:


bobthecow:
Thanks. That's what I get for testing on a couple of versions locally then pushing to master, rather than pushing to a branch 😬

GrahamCampbell:
Lol, it's always the one time you don't check. ;)

PR #613, opened 2020-03-22
Tweaked hoa/console version constraint

GrahamCampbell:
Looks like they are not following semver. `master` is aliased as 3.x, yet they have made breaking changes to method signatures of non-final classes: https://github.com/hoaproject/Console/compare/3.17.05.02...master.

PR #616, opened 2020-03-22
Tweaked hoa/console version constraint

GrahamCampbell:
Replaces https://github.com/bobthecow/psysh/pull/613.


bobthecow:
✅ Confirmed that this works for PHP 5.5+

Issue #572, opened 2019-09-26
Tab Completion of Interfaces


simensen:
Playing around with tab completion related to #539. Another thing I've discovered is you cannot tab complete interfaces, only classes. Would it be appropriate to extend the ClassNameMatcher to include interfaces as well, or to add a new matcher that can be added for people that want it (or by configuration)?

https://github.com/bobthecow/psysh/blob/master/src/TabCompletion/Matcher/ClassNamesMatcher.php#L43

Use case: many DI containers allow you to bind specific implementations to interfaces. It would be great to be able to ask the container for the interface and have it provide the appropriate (wired) implementation.

```
>>> app(\My\Project\CatalogApi::class)
=> My\Project\Adapter\DatabaseCatalogApi {#3541}
```

To be clear, I'm not asking that we allow tab completion to do any sort of interface <-> class mapping, I'm suggesting we allow tab completion interfaces because something else might be able to use that interface to determine the appropriate class behind the scenes. :)

bobthecow:
Yeah, that seems reasonable. I'd feel okay with ClassNamesMatcher matching all class-like things, so interfaces and traits?

PR #617, opened 2020-03-22
Auto-complete interfaces as well as classes

GrahamCampbell:
Closes https://github.com/bobthecow/psysh/issues/572.


bobthecow:
c'mon, at least throw in a single test case ;)

GrahamCampbell:
> c'mon, at least throw in a single test case ;)

Was typing it up already. Am going at one file at a time in the UI. ;)

bobthecow:
Thanks!

PR #618, opened 2020-03-22
Fixed travis build config warnings

GrahamCampbell:
They are also complaining about "skip_cleanup", but their docs don't say how to use the alternative... https://docs.travis-ci.com/user/deployment#uploading-files-and-skip_cleanup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: php
 
-sudo: false
-
-matrix:
+jobs:
     include:
         - php: 5.5.9
           env: 'COMPOSER_FLAGS="--prefer-lowest --prefer-stable"'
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This coverage reporting seems very unreliable. Is it merging all reports, or just the first that finishes?

bobthecow:
Yeah, I don't know what's going on with the coverage report. You might be right that it's only reporting one. When I initially set it up, it definitely included multiple.

bobthecow:
@GrahamCampbell yeah, it seems like it's adding a comment as soon as the first build finishes, even though the overall coverage report merges all of them. It should be fixed with 8201cad924e727cab4e11ecbd32f80cd373fc75e

Issue #621, opened 2020-04-03
Command input without a PTY no longer works in 0.10


phil-s:
There's probably a method that doesn't need ssh, but for instance try this with both 0.9 and 0.10:

    ssh localhost -T -C /path/to/psysh

In 0.9:
```
Psy Shell v0.9.12 (PHP 7.2.24-0ubuntu0.18.04.3 — cli) by Justin Hileman
New version is available (current: v0.9.12, latest: v0.10.2)
42
=> 42
"hello"
=> "hello"
^C
```

In 0.10:
```
Psy Shell v0.10.2 (PHP 7.2.24-0ubuntu0.18.04.3 — cli) by Justin Hileman
42   
"hello"
PHP Parse error: Syntax error, unexpected T_CONSTANT_ENCAPSED_STRING on line 2
Exit:  Exiting main thread
```

My interface connects to the psysh process via a pipe rather than a pty, so this has bitten me in practice, and I've downgraded to 0.9 in the interim.

phil-s:
Ah, so this is the new `--no-interaction` option implicitly kicking in.  If I run `psysh --no-interaction` in a terminal in the normal way, I get the same 'problem' behaviour.

phil-s:
I can't see a config file option for this.  A new config file option which forces the normal "interaction" behaviour would be perfect.

Tangentially, the `--help` text is missing `--no-interaction` and all the other new options listed in the release notes:
```
Psy Shell v0.10.2 (PHP 7.2.24-0ubuntu0.18.04.3 — cli)

Usage:
  psysh [--version] [--help] [files...]

Options:
  --help     -h Display this help message.
  --config   -c Use an alternate PsySH config file location.
  --cwd         Use an alternate working directory.
  --version  -V Display the PsySH version.
  --color       Force colors in output.
  --no-color    Disable colors in output.
```

Perhaps that help could be brought up to date?

bobthecow:
Yeah, something’s not quite right with non-interactive input right now. I’ll dig into it and see what’s up.

In the meantime, I’m pushing a fix for the `--help` output :)

![CC0F4FA4-3234-473A-82BF-61818100A1F5](https://user-images.githubusercontent.com/53660/78382555-e9b3cc80-758b-11ea-9bf1-aeb60e371564.jpeg)

bobthecow:
Looking at your use case, I think we’ll need a flag for “force interactive mode”? I’m assuming you’re piping input as well as output?

[...]

PR #622, opened 2020-04-04
Better plumbing.

bobthecow:
This removes an implicit dependency on side effects from command line options, which was caused by name collisions with the base Console Application.

Things like `--verbose` and `--no-interaction` worked ... kind of. But not always (see #621, for example). And when it did work, it's only because both PsySH and the base Applcation were parsing argv and reading the options passed.

Improved plumbing, specifically:
 - When output decoration isn't explicitly specified, check whether stdout is piped.
 - When the interactive mode isn't explicitly specified, detect whether stdin is interactive.

Improved output:
 - Suppress startup messages when output is piped.
 - Write exceptions to stderr (to make it easier to split them off from desired output).
 - Suppress the little newline marker when `--raw` or piping stdout.

New configuration:
- Add interactive mode configuration, update `psy\info()` output.
- Add verbosity configuration, pass it through to ShellOutput.

Improved configuration:
- Clean up color mode config code.
- Move responsibility for command line input overrides into `Configuration::fromInput`
- Move responsibility for input option declarations into `Configuration::getInputOptions`
- Clean up command-line overrides for configuration.
- Make config errors more consistent.

Drive by:
- Add `--ansi` and `--no-ansi` aliases for color cli options (to match Symfony console and Composer).

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -833,6 +833,8 @@ class Configuration
     public function setOutput(ShellOutput $output)
     {
         $this->output = $output;
+        $this->pipedOutput = null; // Reset cached pipe info
+        $this->applyFormatterStyles();
     }
 
     /**
@@ -848,14 +850,20 @@ class Configuration
     public function getOutput()
     {
         if (!isset($this->output)) {
-            $this->output = new ShellOutput(
+            $this->setOutput(new ShellOutput(
                 OutputInterface::VERBOSITY_NORMAL,
-                $this->getOutputDecorated(),
+                null,
                 null,
                 $this->getPager()
-            );
+            ));
 
-            $this->applyFormatterStyles();
+            // This is racy because `getOutputDecorated` needs access to the
+            // output stream to figure out if it's piped or not, so create it
+            // first, then update after we have a stream.
+            $decorated = $this->getOutputDecorated();
+            if ($decorated !== null) {
+                $this->output->setDecorated($decorated);
+            }
         }
 
         return $this->output;
@@ -870,7 +878,7 @@ class Configuration
     {
         switch ($this->colorMode()) {
             case self::COLOR_MODE_AUTO:
-                return null;
+                return $this->outputIsPiped() ? false : null;
             case self::COLOR_MODE_FORCED:
                 return true;
             case self::COLOR_MODE_DISABLED:
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -34,6 +34,10 @@ class Configuration
     const COLOR_MODE_FORCED   = 'forced';
     const COLOR_MODE_DISABLED = 'disabled';
 
+    const INTERACTIVE_MODE_AUTO     = 'auto';
+    const INTERACTIVE_MODE_FORCED   = 'forced';
+    const INTERACTIVE_MODE_DISABLED = 'disabled';
+
     const VERBOSITY_QUIET        = 'quiet';
     const VERBOSITY_NORMAL       = 'normal';
     const VERBOSITY_VERBOSE      = 'verbose';
@@ -51,6 +55,7 @@ class Configuration
         'forceArrayIndexes',
         'formatterStyles',
         'historySize',
+        'interactiveMode',
         'manualDbFile',
         'pager',
         'prompt',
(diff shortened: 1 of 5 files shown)


test/ShellTest.php
```diff
@@ -384,6 +384,8 @@ public function testWriteStdout()
 
     public function testWriteStdoutWithoutNewline()
     {
+        $this->markTestSkipped('This test won\'t work on CI without overriding pipe detection');
```
GrahamCampbell, line 387: This test is always skipped, even not on the CI?

bobthecow: It'll fail locally if you pipe your phpunit output, too :)

I didn't really want a mock or an override or anything for the piped output detection, but I think I'm going to need it.

bobthecow: It feels like there might be one more configuration option missing. Some way to say "force this to act like output isn't piped", just like we have "force this to decorate output" and "force this to act like input is interactive".

I haven't decided yet whether that's exactly the same as "force this to decorate output", but it might be.

src/Shell.php
```diff
@@ -350,17 +351,18 @@ public function doRun(InputInterface $input, OutputInterface $output)
...
+    private function doInteractiveRun()
     {
         $this->initializeTabCompletion();
         $this->readline->readHistory();
 
-        $this->output->writeln($this->getHeader());
-        $this->writeVersionInfo();
-        $this->writeStartupMessage();
+        // Show a startup message, as long as output isn't piped.
+        if (!$this->config->outputIsPiped()) {
```
bobthecow, line 361: After playing with it a bit, this isn't right. A forced interactive run should have a header, just like an implicitly interactive run. Pushing a fix for this.

bobthecow:
This all feels pretty good, with the exception of the cli option names. I feel like `--interaction` (by analogy to `--no-interaction`) is just as awkward as `--ansi` rather than `--color`. Here's what feels best so far:

```
Psy Shell v0.10.2 (PHP 7.2.28 — cli)

Usage:
  psysh [--version] [--help] [files...]

Options:
  -h, --help            Display this help message.
  -c, --config FILE     Use an alternate PsySH config file location.
      --cwd PATH        Use an alternate working directory.
  -V, --version         Display the PsySH version.
      --color           Force colors in output.
      --no-color        Disable colors in output.
  -i, --interactive     Force PsySH to run in interactive mode.
  -n, --no-interactive  Run PsySH without interactive input. Requires input from stdin.
  -r, --raw-output      Print var_export-style return values (for non-interactive input)
  -q, --quiet           Shhhhhh.
  -v|vv|vvv, --verbose  Increase the verbosity of messages.
```

In addition to those, there are a bunch of aliases; `--[no-]ansi` for `--[no-]color`, `--no-interaction` for `--no-interactive`, to match all the Composer and Symfony console defaults, and `-a` for `-i` just so `psysh -a` does what you'd expect :P

src/Configuration.php
```diff
@@ -134,6 +150,107 @@ public function __construct(array $config = [])
...
+     * @param InputInterface $input
+     *
+     * @return self
+     */
+    public static function fromInput(InputInterface $input)
+    {
+        $config = new self(['configFile' => $input->getOption('config')]);
+
+        // Handle --color and --no-color (and --ansi and --no-ansi aliases)
+        if ($input->getOption('color') || $input->getOption('ansi')) {
```
bobthecow, line 174: I'm feeling really good about this pull request. I think this is my last open question:

Should we use `InputInterface::getOption`?

Using `getOption` like we are here means things will blow up if we're handed input that hasn't been bound to a compatible input options definition. It _is_ more correct, though. We just can't really enforce that people give us good input. 

The more pragmatic approach is probably to use the underlying `hasParameterOption` / `getParameterOption` and do some more of the parsing and interpreting work manually, ourselves. It's not as clean, and not as guaranteed-correct, but it also doesn't rely on callers having already bound compatible options. 

Either way, we'd still continue binding the input properly in `bin/psysh` when _we_ call it. We'd just be more forgiving of other callers not doing the same.

I'm leaning towards the second option.

@GrahamCampbell what do you think? This is somewhat relevant to https://github.com/laravel/tinker/pull/98 🙂

bobthecow: K. I wen't with a third option.

I've got something that works either way, though definitely better with bound input. See 3c9b112.

It's a bit of a mess, but probably acceptably tucked away from end users that it won't be an issue?

PR #628, opened 2020-04-09
Fixed doc rendering where there is no description

### LINKED ISSUE #627: Double bug report
image

Bug 1: The doc command crashes as per the above screenshot (first reported at https://github.com/laravel/tinker/issues/99).

Bug 2: I ran with `-vvv` but the stacktrace still didn't get shown. Symfony's error handler shows the stacktraces for internal errors by default if I pass `-v` or higher. Maybe we should do the same?

GrahamCampbell:
Closes #627.


bobthecow:
Thank you.

PR #630, opened 2020-04-10
Added a way to add loop listeners

GrahamCampbell:


bobthecow:
Why? 

🙂

GrahamCampbell:
https://github.com/laravel/tinker/pull/103 ;)

GrahamCampbell:
Currently it's not easy for someone to add an event listener, once the shell has already been created, and adding the listeners afterwards is definitely going to be the best DX for laravel tinker users.

GrahamCampbell:
Original motivation at https://github.com/laravel/tinker/pull/101.

bobthecow:
It’s come up occasionally. I’m just not convinced opening up loop listeners as a public api is worth the complication. I’m interested in justification for the feature, not just a pull request to implement it.

GrahamCampbell:
The motivation @NiclasvanEyk gave in laravel/tinker#101 was:

> This enables third-parties to react to commands executed during tinker sessions.
> The concrete example that motivated this PR was Laravels telescope package.
> Executed queries, logs or model-events will not be recorded by telescope, as
> the application will not shut down while inside the tinker-loop.

GrahamCampbell:
That is, they were interested in simulating a (partial) application shutdown after each tinker command.

GrahamCampbell:
What's the status here? If this is not wanted, I'll happily close the PR. If changes are required, I can make those too. :)

paras-malhotra:
@bobthecow any chance we could re-open this? The issue at hand still remains.

We're unable to simulate an application shutdown after each tinker command. Packages such as [Laravel Telescope](https://github.com/laravel/telescope), which are great for debugging, simply don't behave the same way in Tinker and that defeats the purpose of the debugging package really.

If there's a way we could do this without this PR, I'd be happy to work on it but I think this would be the cleanest approach.

paras-malhotra:
@bobthecow we tried fixing this at our own end on Tinker but unfortunately, the loop listeners variable is private and not protected, so we're pretty much stuck on this issue. Unless we merge this, it cannot be solved to the best of my knowledge.

What concerns do you have if we have a way to add loop listeners? Is there an alternative that you could suggest?

NiclasvanEyk:
@paras-malhotra If you really want the functionality to execute logic after each loop, couldn't you just maintain your own afterLoopListeners and override [`Shell#afterLoop`](https://github.com/bobthecow/psysh/blob/2e6b273b4af2070913dfe37f6000e7392b0e52cd/src/Shell.php#L613) to also execute *your* custom listeners?

paras-malhotra:
@NiclasvanEyk, yep we could do that and possibly maintain our own additional loop listeners that we call after calling `parent::afterLoop`. Thanks for the suggestion. 

I was just hoping if we could merge this in and it would be so much cleaner! If Justin doesn't reply, we can probably try that approach.

PR #631, opened 2020-04-10
Formatter classes to … just functions?

bobthecow:
[Per a comment earlier today…](https://github.com/bobthecow/psysh/commit/6b3bfefe704397259702573d194a4eef1317934b#commitcomment-38417314)

> These would be functions if PHP supported function autoloading. I'm half tempted to make them functions anyway and just lean on Composer a bit more :)

Is this a terrible idea? It feels a lot cleaner to me.


GrahamCampbell:
The problem with functions like this is they don't play well with global installs of packages along slide local versions. For this reason, Guzzle is refactoring away from all functions to static class methods. https://github.com/guzzle/guzzle/pull/2546 https://github.com/guzzle/guzzle/issues/2511

GrahamCampbell:
If someone has a different version of psysh installed globally to a local install, then composer will definitely load the wrong functions file for one of them...

GrahamCampbell:
If those functions are different or don't exist in one file, then you run into issues, like in Guzzle.

bobthecow:
Globally-installed psysh works around exactly these issues by always deferring to local.

GrahamCampbell:
How does the local install stop composer from autoloading the global functions files first?

GrahamCampbell:
I think that was the exact bug that guzzle was having?

bobthecow:
The global `psysh` binary doesn't use the global composer installation at all if you have something installed locally. It loads the local one:

Screen Shot 2020-05-02 at 12 29 42 AM

Unless I'm misunderstanding the failure mode, you shouldn't ever hit any globally installed PsySH code if you're running a local installed version.

PR #633, opened 2020-04-30
Add dbg constant

kapitanluffy:
Add a customizable constant to be used instead of a namespaced function

```php
// in config.php
return [
    'debugConstant' => 'DBG' // "\Psy\sh" by default
];
```

```php
// implementation
$foo = "change me";

eval(DBG);
echo $foo;
```

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -396,7 +396,7 @@ EOL;
     }
 }
 
-/**
+/*
  * Use `eval(dbg)` because `eval(\Psy\sh());` is too long
  */
 if (!defined('dbg') && function_exists('Psy\\sh')) {
--- a/src/functions.php
+++ b/src/functions.php
@@ -400,5 +400,6 @@ EOL;
  * Use `eval(dbg)` because `eval(\Psy\sh());` is too long
  */
 if (!defined('dbg') && function_exists('Psy\\sh')) {
-    define('dbg', \Psy\sh(), true);
+    $constant = (new Configuration())->getDebugConstant();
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
👎 here, for invading the global namespace.

kapitanluffy:
> 👎 here, for invading the global namespace.

Addressed it by using a namespaced constant by default.

src/functions.php
```diff
@@ -395,3 +395,11 @@ function bin()
         };
     }
 }
+
+/*
+ * Use `eval(sh)` because `eval(\Psy\sh());` is too long
+ */
+$constant = (new Configuration())->getDebugConstant();
```
bobthecow, line 402: Using `Configuration` like this is problematic. There's no guarantee that the configuration you get instantiating it this way is actually the one in use in the eventual PsySH session (for example, [if the config file is overridden via the `--config` flag](https://github.com/bobthecow/psysh/blob/master/src/Configuration.php#L173)).

Even if it does end up with the right config file, adding this in `src/functions.php` (or any other non-autoloaded PHP file) isn't great. It's run on every request to every webserver and every CLI app with PsySH installed as a dependency, regardless of whether someone intends to use it or not.

The current approach is almost a no-op until someone actually calls `Psy\sh()` which kicks off a bunch of autoloading and object instantiating. I don't think we can justify the overhead just to save people a few characters, _especially_ since the default is just `eval(Psy\sh())` to `eval(Psy\sh)`.

PR #642, opened 2020-07-01
Make error message about directory not writable more clear

shapito27:
Several times faced with error 
Screenshot from 2020-07-01 11-27-34

when work inside docker and have not enough rights.
And this message made me confused, because i tried create this file  _/var/www/.config/psysh_ and it still didn't work.
I think will be better if add word **directory**


bobthecow:
Thanks!

PR #646, opened 2020-08-02
Updates for the new Symfony CS

GrahamCampbell:
This PR is a follow-up to an email conversation RE changes to the Symfony preset (following changes to the Symfony style guide itself).


GrahamCampbell:
I think these resent test failures are likely related to changes in the php-parser package. I've not looked in detail, but I suspect the issue started around PHP-Parser 4.6 or 4.7.

bobthecow:
The current failures are symfony/symfony#37674, a regression in VarDumper v5.1.3. It should be fixed now, but apparently not released?

.styleci.yml
```diff
@@ -1,29 +1,27 @@
...
+  - native_constant_invocation_symfony
+  - native_function_invocation_symfony
+  - no_superfluous_phpdoc_tags_symfony
+  - pow_to_exponentiation
   - pre_increment
   - unalign_double_arrow
   - unalign_equals
   - yoda_style
-  - property_separation
-  - const_separation
```
bobthecow, line 22: are there any more overrides we should remove with this new batch of preset changes?

GrahamCampbell: No. If there were, StyleCI would fail loudly that a fixer that was not enabled by the preset cannot be disabled. :)

bobthecow: oh, no, i meant whether we should remove any overrides to get closer to the base preset, not to keep styleci from whining 😉

GrahamCampbell: Oh, I guess pre_increment could go. I don't think there'd be too many changes from that. The others will cause quite a lot of changes, however. There was also the discussion we had about the slashes on internal constants and functions. You could un-disable the two _symfony fixers listed here, or keep them disabled and enable the full versions.

bobthecow: let's leave `pre_increment`. that, along with `yoda_style`, is one of those things where symphony's coding standard makes you think about things backwards in the name of (questionable) protection from yourself.

looking at overlap between laravel, symfony, and styleci's "recommended", it seems like i'm pretty clearly in the minority for `concat_without_spaces`, `method_argument_space` and `unalign_equals`. we can remove those overrides 😢

bobthecow: it might be worth switching the big ones (spaces and alignment changes) in their own diff to keep things cleaner.

PR #654, opened 2020-09-24
Psy\debug: Option to suppress printing of context

sbusch:
Added an option for `Psy\debug` to suppress printing of context (via `whereami` command). Default behaviour stays to print context.

Background: I implemented a simple shell script to boot PsySH with our own variables and found it quite irritating - especially for other developers using that script - that parts of it were printed before getting a prompt.

Would be glad if you could merge this!

Edit: typos

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -71,8 +71,8 @@ if (!\function_exists('Psy\\debug')) {
      *         }
      *     }
      *
-     * @param array         $vars   Scope variables from the calling context (default: [])
-     * @param object|string $bindTo Bound object ($this) or class (self) value for the shell
+     * @param array         $vars            Scope variables from the calling context (default: [])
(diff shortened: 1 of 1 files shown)


bobthecow:
The `psy\debug` function is meant for YOLO inline usage—interactive testing like you'd use `debug` for in javascript. Anything more complicated should probably use the proper API, which is made for this sort of thing. It doesn't automatically add a `whereami` call:

```php
$shell = new \Psy\Shell();
$shell->run();
```

This API gives you an opportunity to tweak configuration, add extra commands, set up initial variables, etc before executing. It's what Laravel Tinker and our own `psysh` binary use internally.

sbusch:
Thanks for your comment! I just thought it could be a useful shortcut for a possibly common use case, but I have no problem with you rejecting this PR, so I'm closing it... still loving PsySH ;-)

Issue #657, opened 2020-10-29
"hist" command does not show history


tyler36:
There is a good chance I'm just doing something stoopid here but ...

1. Open powershell prompt
```
❯ psysh
Psy Shell v0.10.4 (PHP 7.3.18 — cli) by Justin Hileman
>>> 1 + 2
=> 3
>>> $a = 'apple'
=> "apple"
>>> hist
>>>                                                                                                                 
```

Am I missing something or should it be displaying as shown (https://github.com/bobthecow/psysh/wiki/History)? 
```
>>> hist
0: $a = 'apple'
1: 1 + 2
>>>
```

The history file appears configured:
```
>>> Psy\info()
...
     "history" => [
       "history file" => "~/AppData/Roaming/PsySH/psysh_history",
       "history size" => null,
       "erase duplicates" => null,
     ],
...
```

If I check the file, it contains entries
```
1 + 2
$a = 'apple'
hist
Psy\info()
```

I can navigate through the histroy the up/down arrow keys. ```hist --clear``` also empties the file. 
I just can't get Psysh to display the histroy.

What am I missing?

Psysh: 0.10.4
PHP: 7.3.18
OS: Win10

bobthecow:
Can you share the rest of `psy\info()`?

tyler36:
Thanks for the quick reply:
```
>>> psy\info()
=> [
     "PsySH version" => "v0.10.4",
     "PHP version" => "7.3.18",
     "OS" => "WINNT",
     "default includes" => [],
     "require semicolons" => false,
     "error logging level" => 32767,
     "config file" => [
       "default config file" => null,
       "local config file" => null,
       "PSYSH_CONFIG env" => false,
     ],
     "loop listeners" => [],
     "commands" => [
       "help" => "Psy\Command\HelpCommand",
       "?" => "Psy\Command\HelpCommand",
       "ls" => "Psy\Command\ListCommand",
       "dir" => "Psy\Command\ListCommand",
       "dump" => "Psy\Command\DumpCommand",
       "doc" => "Psy\Command\DocCommand",
       "rtfm" => "Psy\Command\DocCommand",
       "man" => "Psy\Command\DocCommand",
       "show" => "Psy\Command\ShowCommand",
       "wtf" => "Psy\Command\WtfCommand",
       "last-exception" => "Psy\Command\WtfCommand",
       "wtf?" => "Psy\Command\WtfCommand",
       "whereami" => "Psy\Command\WhereamiCommand",
[...]

PR #658, opened 2020-10-29
🛠️ hack-around fix for "hist --clear" command

tyler36:
A "bad" workaround for #657 

But it might be a start for someone with a better solution.

Test system 1:
OS: Win10
PHP: 7.4.6 / 7.3.18
Psysh: 0.10.4

### DIFF
--- a/src/Command/HistoryCommand.php
+++ b/src/Command/HistoryCommand.php
@@ -31,7 +31,7 @@ class HistoryCommand extends Command
 
     /**
      * Required file header for 'readline_read_history' command
-     * https://www.php.net/manual/en/function.readline-read-history.php
+     * https://www.php.net/manual/en/function.readline-read-history.php .
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


bobthecow:
We're actually enforcing that ourselves, to keep us from incorrectly interpreting a non-history file as history:

https://github.com/bobthecow/psysh/blob/master/src/Readline/Libedit.php#L53-L56

Per the spec, it's _supposed_ to have that signature. I'm a bit grumpy that it doesn't.

But if we can't guarantee it's there, we could make that conditional, and just… let it fail if something goes wrong? ¯\\\_(ツ)\_/¯ 

```php
// remove history signature if it exists
if ($history[0] === '_HiStOrY_V2_') {
    \array_shift($history);
}
```

tyler36:
Sounds good to me.  

I bounce between Laravel & Drupal projects and both have their own wrappers around psysh ( ```drush``` & ```tinker```).  Having the (searchable) history feature would be amazing. 

Please let me know if there is anything I can do to help.

bobthecow:
Awesome! Closing in favor of the fix in b04ee23

PR #663, opened 2020-12-05
Don't bother testing on broken versions of HHVM

GrahamCampbell:
Just a waste of CO2 I think, to test versions that never worked, and that nobody is going to fix (I assume).


bobthecow:
Yeah. That's fair. I figured I'd include them in case I was missing something. But it seems like it's just not compatible ¯\_(ツ)_/¯

PR #667, opened 2021-02-10
add support for runkit7 < 4.0.0a1

maxime-aknin:
Enable partial php 7 support for [runkit7](https://www.php.net/manual/en/book.runkit7.php) before version 4.0.0a1 where `runkit_import` function was removed.


bobthecow:
Thanks! Mostly-working runkit7 is better than no runkit at all, I guess 🙂

PR #676, opened 2021-05-16
CS fixes for upcoming StyleCI release

GrahamCampbell:
This repo is included in StyleCI's internal tests to check we don't break the repo. The upcoming deploy on Sunday 16th will make the fixes included in this PR. I am sending this PR over ahead of time so it's clear where these fixes came from, since you made no config changes, but the fixing output changed - this is because we're treating this as a bug fix. :)


bobthecow:
Thank you!

PR #693, opened 2021-11-30
Dropped super old deps

GrahamCampbell:
PHP-Parser 2.0.0 only supported PHP 7.0, and nothing greater, and support for symfony console 2 was already dropped, so ditching symfony var-dumper 2 seems fair, too. Moving the min PHP version to 7.0.8 also enables the PHP 7.0 compatable phar build to use the latest symfony 3.x components.


bobthecow:
Thanks! I'd leave the PHP-Parser 2.x in there as long as we still support PHP 7.0, because it's no extra work to support and isn't holding us back. Other than that, 👍

bobthecow:
ehhhh poked at it a bit, and dropping 2.x reduces a bunch of the noisy fake failures in our `--prefer-lowest` tests, so i guess that's enough of a support cost ¯\_(ツ)_/¯

bobthecow:
Thanks again :)

PR #694, opened 2021-12-13
Set allow-plugins for Composer 2.2

villfa:
This new configuration has been introduced with Composer 2.2: https://github.com/composer/composer/releases/tag/2.2.0-RC1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,11 @@
         }
     },
     "bin": ["bin/psysh"],
+    "config": {
+        "allow-plugins": {
+            "bamarni/composer-bin-plugin": true
+        }
+    },
     "extra": {
         "branch-alias": {
             "dev-main": "0.10.x-dev"


GrahamCampbell:
It may also be necessary to patch this on the 0.10 branch. :)

villfa:
The PR has been updated to target the 0.10 branch.

bobthecow:
Making sure I understand correctly: this is a flag for development _on this project_, and won't affect projects that list PsySH as a dependency, right? It's just for our ability to develop using this Composer plugin?

GrahamCampbell:
Assuming bamarni is used by github actions, it'll also stop things breaking on the upgrade to composer 2.2.

villfa:
> Making sure I understand correctly: this is a flag for development _on this project_, and won't affect projects that list PsySH as a dependency, right? It's just for our ability to develop using this Composer plugin?

I confirm it will affect only this project.

More details here: https://getcomposer.org/doc/06-config.md#allow-plugins

bobthecow:
Right, I read the docs. But Composer doesn't always do a great job of specifying local config vs config that propagates to projects that pull in a dependency so I wasn't entirely sure what the scope of the change was.

bobthecow:
Thank you. Merged into both `0.10` and `main` branches.

PR #696, opened 2021-12-15
Improve PHPUnit assertions

peter279k:
# Changed log

- Using the `assertSame` to make assertion equals strict.


bobthecow:
Great, thank you!

PR #697, opened 2022-01-03
fix nullable params for getHistorySlice and allow for custom historyFile

tgooren:
getHistorySlice was broken due to the added typehints (only one of the 3 params will be a string). I've also added `historyFile` to the available options in the configuration to allow for supplying a custom HistoryFile location in the config file.

### DIFF
--- a/src/Command/HistoryCommand.php
+++ b/src/Command/HistoryCommand.php
@@ -182,7 +182,7 @@ HELP
      *
      * @return array A slice of history
      */
-    private function getHistorySlice(?string $show, ?string $head, ?string $tail): array
+    private function getHistorySlice($show, $head, $tail): array
     {
         $history = $this->readline->listHistory();
 
--- a/src/Command/HistoryCommand.php
+++ b/src/Command/HistoryCommand.php
@@ -176,13 +176,13 @@ HELP
     /**
      * Retrieve a slice of the readline history.
      *
-     * @param string $show
-     * @param string $head
-     * @param string $tail
+     * @param string|null $show
+     * @param string|null $head
+     * @param string|null $tail
      *
-     * @return array A slilce of history
+     * @return array A slice of history
      */
-    private function getHistorySlice(string $show, string $head, string $tail): array
+    private function getHistorySlice(?string $show, ?string $head, ?string $tail): array
     {
         $history = $this->readline->listHistory();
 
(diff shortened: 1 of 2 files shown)


src/Command/HistoryCommand.php
```diff
@@ -176,13 +176,13 @@ private function extractRange(string $range): array
...
-     * @param string $tail
+     * @param string|null $show
+     * @param string|null $head
+     * @param string|null $tail
      *
-     * @return array A slilce of history
+     * @return array A slice of history
      */
-    private function getHistorySlice(string $show, string $head, string $tail): array
+    private function getHistorySlice(?string $show, ?string $head, ?string $tail): array
```
bobthecow, line 185: PsySH supports PHP 7.0, so we need to remove the type hints rather than marking them nullable.

src/Command/HistoryCommand.php
```diff
@@ -176,13 +176,13 @@ private function extractRange(string $range): array
...
      *
-     * @param string $show
-     * @param string $head
-     * @param string $tail
+     * @param string|null $show
+     * @param string|null $head
+     * @param string|null $tail
      *
-     * @return array A slilce of history
+     * @return array A slice of history
```
bobthecow, line 183: 🙃

src/Configuration.php
```diff
@@ -56,6 +56,7 @@ class Configuration
         'errorLoggingLevel',
         'forceArrayIndexes',
         'formatterStyles',
+        'historyFile',
```
bobthecow, line 59: Interesting that we didn't already support this.

It seems to be the only non-service settable config that's not included in the available options, and it predates the ability to set them this way, so maybe it just slipped through the cracks? I can't think of any reason not to support it ¯\\\_(ツ)\_/¯

bobthecow (changes requested):

src/Configuration.php
tgooren: it's something we had in an internal fork for years, Wondered why we never made a pull request for it :)

src/Command/HistoryCommand.php
tgooren: I'll remove the typehints then 👍

tgooren: I've removed the typehints. I left the docblock update, to avoid someone accidently reintroducing the same typehints in the future

bobthecow (approved):

bobthecow:
Thank you!

PR #716, opened 2022-05-09
Suppress WARNING for TERM=dumb

zonuexe:
The `less` command outputs a WARNING message in the environment of `TERM=dumb`, so suppress this.

スクリーンショット 2022-05-10 4 00 15

I'm going to fix some issues in the dumb environment. refs https://github.com/emacs-php/psysh.el/issues/8

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -1209,6 +1209,10 @@ class Configuration
     public function getPager()
     {
         if (!isset($this->pager) && $this->usePcntl()) {
+            if (\getenv('TERM') === 'dumb') {
+                return false;
+            }
+
             if ($pager = \ini_get('cli.pager')) {
                 // use the default pager
                 $this->pager = $pager;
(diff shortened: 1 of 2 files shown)


zonuexe:
I think this PR is already mergeable, but I don't think it's necessary to rush to release a new version as other issues need to be fixed.

I will submit another PR when new issues are identified regarding the compatibility of the dumb environment.

src/Configuration.php
```diff
@@ -1198,10 +1198,14 @@ public function setPager($pager)
...
-     * @return string|OutputPager
+     * @return string|OutputPager|null
      */
     public function getPager()
     {
+        if (\getenv('TERM') === 'dumb') {
+            return null;
+        }
+
         if (!isset($this->pager) && $this->usePcntl()) {
```
bobthecow, line 1209: Thanks for this!

I was thinking it might make more sense to put the check for a dumb terminal here, i.e. if a pager isn't explicitly defined, don't define one. That way we leave room for users who intentionally add a pager via config, but don't force a pager when it doesn't make sense.

zonuexe: @bobthecow Thank you for your review. I made it return `false` instead of `null` that same as  860c4819a09a43712eaa88fee2348e6d5d094ffe .

bobthecow:
Thank you!

PR #719, opened 2022-05-22
chore(deps): Included dependency review

naveensrinivasan:
> Dependency Review GitHub Action in your repository to enforce dependency
> reviews on your pull requests.
> The action scans for vulnerable versions of dependencies introduced by package version
> changes in pull requests,
> and warns you about the associated security vulnerabilities.
> This gives you better visibility of what's changing in a pull request,
> and helps prevent vulnerabilities being added to your repository.

https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/depsreview.yaml
+++ b/.github/workflows/depsreview.yaml
@@ -1,0 +1,14 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v1


bobthecow:
My initial thought is that things like this make sense for projects, but less so for libraries.

PsySH intentionally takes very few dependencies, doesn't have lockfiles, and keeps the version ranges as wide as possible, to let project owners and end users make the decisions that make the most sense for them. We had six external dependencies seven years ago, then five dependencies two years ago, dropped down to four a year ago, and to three at the start of this month. At this rate we'll run out of dependencies soon!

Honestly, pull requests that try to add a dependency, especially one pinned to a specific version, would probably be rejected for introducing a dependency :)

naveensrinivasan:
> My initial thought is that things like this make sense for projects, but less so for libraries.
> 
> PsySH intentionally takes very few dependencies, doesn't have lockfiles, and keeps the version ranges as wide as possible, to let project owners and end users make the decisions that make the most sense for them. We had six external dependencies seven years ago, then five dependencies two years ago, dropped down to four a year ago, and to three at the start of this month. At this rate we'll run out of dependencies soon!
> 
> Honestly, pull requests that try to add a dependency, especially one pinned to a specific version, would probably be rejected for introducing a dependency :)

Sounds good. I can close this. Thanks

bobthecow:
This did remind me of something I'd been meaning to do. I just added bf05703 to ensure that our pre-built phar releases don't include known vulnerabilities.

PR #726, opened 2022-06-22
adding :q shortcut for VIM users

tahashieenavaz:
A true VIM user will think of this command when leaving any sort of window in OS. 
It would be great to have it here too.


bobthecow:
Unfortunately the command system doesn't support `:` :( 

> PHP Fatal error:  Uncaught Symfony\Component\Console\Exception\InvalidArgumentException: Command name ":q" is invalid.

bobthecow:
I'm closing this for now. Feel free to reopen with a working implementation :)

Issue #729, opened 2022-07-01
Pager process crashes on macOS


zonuexe:
An error occurs when outputting a long value to the pager on macOS.
No value can be output once the process crashes.

https://user-images.githubusercontent.com/822086/176931967-6e17cb49-008e-49af-a688-9442489579fc.mov

From what I've looked at, `range(1, 3563)` doesn't cause any problems and `range(1, 3564)` raises an exception.

This problem does not occur on Linux.

bobthecow:
Yeah, that's weird. Your workaround helps it successfully recover, though, so that's great!

At least on my machine, after your change it only fails when trying to close the pager, so at that point it's … probably fine? The workaround makes it not terrible :)

PR #730, opened 2022-07-01
Close the pipe if the pager process goes wrong

zonuexe:
refs https://github.com/bobthecow/psysh/issues/729

By closing the process after the write fails, the new value will be output.


bobthecow:
Thanks!

Issue #732, opened 2022-07-11
Activate auto-completion for non-GNU Readlines


zonuexe:
The current AutoCompleter enables `readline_completion_function()` directly, but this completion mechanism can be provided in addition to the native `readline()` function. So is the `UserlandReadline`, which is based on Hoa, for example.

<details><summary><s>deleted</s></summary><del>

The current AutoCompleter enables `readline_completion_function()` directly, but this completion mechanism can be provided in addition to the native `readline()` function. So is the `UserlandReadline`, which is based on Hoa, for example.

The `nodejs-repl.el` suggested by @takeokunn seems to do exactly that.

</del></details>

zonuexe:
My next goal is to communicate with each other for input completion with Emacs REPL, which requires proper separation of AutoCompleter.

refs https://github.com/emacs-php/psysh.el/issues/9

phil-s:
I have an old not-forgotten-but-languishing-in-the-backlog PR for providing completion to non-readline cases (also for Emacs, as it happens).
 
- https://github.com/bobthecow/psysh/issues/555
- https://github.com/bobthecow/psysh/pull/649

I've only spent a matter of seconds skimming over #733 and the code seems to deal with Hoa (which my code does not), but nevertheless I imagine there would be some overlap between the two.

At the time I'd stalled on reworking some of the completion code and implementing tests (as an outcome of my work was the discovery that there was some weirdness in the existing behaviour, and very little test coverage).  I believe I still have unpushed WIP for that (which definitely needed more work).

bobthecow:
@phil-s i've actually started working on a rework of some readline stuff myself, and i'd welcome your feedback/help on it once i get it published. i think the current Hoa thing is a dead end, and we can do a lot better!

zonuexe:
I haven't been able to work on these things for a while, but the approach away from Hoa looks very promising (I was thinking about working on a PHP port of [Ruby's Reline](https://github.com/ruby/reline) sometime in my spare time).

bobthecow:
let me get my feature branch cleaned up and published. i've done a lot of work on it already, and i'd appreciate help from both of you :)

bobthecow:
@phil-s @zonuexe check out #914 when you get a chance?

bobthecow:
This is implemented [in the latest release](https://github.com/bobthecow/psysh/releases/tag/v0.12.21) (via the `--experimental-readline` flag). Try it out and let me know how it goes!

PR #733, opened 2022-07-11
Implemented input completion for Userland

zonuexe:
refs https://github.com/bobthecow/psysh/issues/732

### DIFF
--- a/src/Readline/Hoa/Autocompleter.php
+++ b/src/Readline/Hoa/Autocompleter.php
@@ -46,8 +46,11 @@ interface Autocompleter
     /**
      * Complete a word.
      * Returns null for no word, a full-word or an array of full-words.
+     *
+     * @param array{line_buffer: string} $info A subset of {@see readline_info()}'s return value.
+     * @see https://www.php.net/readline_info
      */
-    public function complete(&$prefix);
+    public function complete(string $prefix, int $index, array $info);
 
     /**
      * Get definition of a word.
--- a/src/Readline/HoaAutocompleterAdapter.php
+++ b/src/Readline/HoaAutocompleterAdapter.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Psy\Readline;
+
+use Psy\TabCompletion\AutoCompleter;
+use Psy\Readline\Hoa\Autocompleter as HoaAutocompleter;
+
+class HoaAutocompleterAdapter implements HoaAutocompleter
+{
+    /** @var AutoCompleter */
+    private $autoCompleter;
+
+    public function __construct(AutoCompleter $autoCompleter)
+    {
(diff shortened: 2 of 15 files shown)


src/Readline/HoaAutocompleterAdapter.php
```diff
@@ -0,0 +1,27 @@
...
+    public function complete(string $prefix, int $index, array $info)
+    {
+        return $this->autoCompleter->complete($prefix, $index, $info);
+    }
+
+    public function getWordDefinition(): string
+    {
+        return '.';
+    }
+}
```
zonuexe, line 27: I left this adapter for interoperability with existing Hoa completions, but you can also remove `HoaAutocompleter` and make it directly dependent on `Psy\TabCompletion\AutoCompleter`. What do you think?

src/Readline/Readline.php
```diff
@@ -80,4 +82,14 @@ public function redisplay();
...
+
+    /**
+     * Activete auto completer for tab completion.
+     */
+    public function activateAutoCompleter(AutoCompleter $autoCompleter): void;
+
+    /**
+     * Deactivete auto completer for tab completion.
+     */
+    public function deactivateAutoCompleter(): void;
```
zonuexe, line 94: Do nothing for deactivate except native `readline()`.

src/Readline/Readline.php
```diff
@@ -80,4 +82,14 @@ public function redisplay();
      * @return bool Success
      */
     public function writeHistory(): bool;
+
+    /**
+     * Activete auto completer for tab completion.
+     */
+    public function activateAutoCompleter(AutoCompleter $autoCompleter): void;
```
bobthecow, line 89: let's make these two a second interface so we can maintain backwards compatibility for existing Readline API.

bobthecow (changes requested):
Realistically we're a while from a release where we can introduce backwards compatibility breaks.

For better or worse, we need to maintain compatibility with the current `AutoCompleter` and `Readline` interfaces (though i'm not worried about anything in `Psy\Readline\Hoa\*`, that's an implementation detail not a public API 😛)

src/Readline/Hoa/Autocompleter.php
```diff
@@ -46,8 +46,11 @@ interface Autocompleter
     /**
      * Complete a word.
      * Returns null for no word, a full-word or an array of full-words.
+     *
+     * @param array{line_buffer: string} $info A subset of {@see readline_info()}'s return value.
+     * @see https://www.php.net/readline_info
      */
-    public function complete(&$prefix);
+    public function complete(string $prefix, int $index, array $info);
```
bobthecow, line 53: these were passing by reference because they changed `$prefix` intentionally so the caller could check prefix length.

Issue #722, opened 2022-05-30
bug: Laravel tinker won't start because of a missing Termfile?


sts-ryan-holton:
I've just tried to start tinker in my Laravel 8 project, I'm working on a Windows 10 PC with Mamp, I'm running PHP 7.4 and when running `php artisan tinker` I get the following error:

> Terminfo file does not exist

I originally posted this on a Stackoverflow and it seems that this is a recent change of this library that has been made to look for a terminfo file?

I've tried manually installing the `dev` branch into my project as a dependency with as part of a [suggestion](https://stackoverflow.com/a/72409712/9982090):

```
composer require psy/psysh:@dev
```

But this doesn't resolve my issue and I continue to experience the same error. Furthermore, I have tried changing PHP versions from 7.4 to 7.3 and still get the same error, I've also checked that I have the `php_pdo_sqlite` extension installed and active and I do.
![7G5sX](https://user-images.githubusercontent.com/25177538/171046375-c32b9484-1ce7-4bb9-b3e7-5420c0aef921.png)

My production site on a Cent OS 8 machine running PHP 7.4 doesn't experience this error.

bobthecow:
This exception _should_ be prevented by a recent change in dev. Can you confirm that the `isSupported` method in `Psy\Readline\Userland` reads:

```php
        return HoaUstring::checkMbString() && HoaConsoleTput::getTerminfo() !== '';
```

sts-ryan-holton:
@bobthecow When I run:

```
composer require psy/psysh:@dev
```

The following is added to my composer.json file:

```
"psy/psysh": "@dev",
```

This is the output from my command line
![Capture](https://user-images.githubusercontent.com/25177538/171235375-901fe108-3f32-4933-9889-acf6b428cbf0.PNG)

And then when looking at the location you suggested, I do not see that line. The method still looks like:

```php
/**
 * @return bool
 */
public static function isSupported(): bool
{
    return HoaUstring::checkMbString();
}
```

If I manually add the last bit, then I get this error:

> file_exists(): Unable to find the wrapper "hoa" - did you forget to enable it when you configured PHP?

Still running PHP 7.4

I also tried doing:

```
composer require psy/psysh:@develop

and

composer require psy/psysh:@main
```

still no luck

bobthecow:
yeah, you're definitely not on the most recent psysh. what does `composer info` show?

sts-ryan-holton:
[...]

PR #734, opened 2022-07-26
Bootstrap Hoa stream wrapper during Userland capability check (isSupported)

daniser:
Fix for #722.

### DIFF
--- a/src/Readline/Userland.php
+++ b/src/Readline/Userland.php
@@ -40,6 +40,8 @@ class Userland implements Readline
      */
     public static function isSupported(): bool
     {
+        static::bootstrapHoa();
+
         return HoaUstring::checkMbString() && HoaConsoleTput::isSupported();
     }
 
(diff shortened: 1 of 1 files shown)


src/Readline/Userland.php
```diff
@@ -40,6 +40,8 @@ class Userland implements Readline
      */
     public static function isSupported(): bool
     {
+        static::bootstrapHoa();
```
bobthecow, line 43: i _think_ `bootstrapHoa` doesn't need to bind terminal resize in this case (because we're not trying to use it, we're just trying to see if we _can_ use it).

maybe we make this optional when bootstrapping?

```
\class_exists('Psy\Readline\Hoa\ConsoleWindow');
```

daniser: > i _think_ `bootstrapHoa` doesn't need to bind terminal resize in this case (because we're not trying to use it, we're just trying to see if we _can_ use it).
> 
> maybe we make this optional when bootstrapping?
> 
> ```
> \class_exists('Psy\Readline\Hoa\ConsoleWindow');
> ```

Check out my latest changes. Is this what you meant?

bobthecow: Yep, thanks!

bobthecow:
Thank you!

PR #737, opened 2022-07-29
Improves console output

nunomaduro:
After https://github.com/laravel/tinker/pull/152, as requested, I've made this pull request that updates the console output of the shell class. Here are the changes, and feel free to "check" the ones you agree with, and leave uncheck the changes you think still need some discussion.

- [x] As requested, we've left the library's author name unchanged, and simply updated the color from "blue" to "cyan".
- [ ] Next, the `PROMPT`, `BUFF_PROMPT`, `REPLAY, and `RETVAL`, have been simplified to a single `char`. Also, the `RETVAL` is displayed on `gray`, so the output result gets highlighted.
- [ ] After, when errors/warnings occur, I've added margin-top and margin-bottom ( a new line ), so the terminal has some space to breathe.
- [x] Also, the "Goodbye" message, or any message coming from "BreakException" are now displayed as regular "INFO" information, as they are not errors.
- [x] Next, I've displayed the errors differently, using a "CLASS_NAME" followed by the original message for regular exceptions, and using "ERROR_TYPE" followed by the original message for PHP errors/warnings.
- [x] I've removed the "in Psy Shell code" wording, as it doesn't add anything.
- [x] Finally, I've ensured error/warnings messages **always** end with a "dot", and contain relative paths instead of absolute paths.

Here is the described changes, but comparing the before vs after:

### DIFF
--- a/src/Output/ShellOutput.php
+++ b/src/Output/ShellOutput.php
@@ -218,8 +218,6 @@ class ShellOutput extends ConsoleOutput
         $formatter->setStyle('inline_html', new OutputFormatterStyle('cyan'));
     }
 
-
-
     /**
      * Checks if the "gray" color exists on the output.
      *
--- a/src/Output/ShellOutput.php
+++ b/src/Output/ShellOutput.php
@@ -176,13 +176,17 @@ class ShellOutput extends ConsoleOutput
     {
         $formatter = $this->getFormatter();
         $errorFormatter = $this->getErrorOutput()->getFormatter();
+        $grayExists = $this->grayExists();
 
+        $formatter->setStyle('info', new OutputFormatterStyle('white', 'blue', ['bold']));
+        $errorFormatter->setStyle('info', new OutputFormatterStyle('white', 'blue', ['bold']));
         $formatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
         $errorFormatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
         $formatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
         $errorFormatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
+        $formatter->setStyle('whisper', new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
+        $errorFormatter->setStyle('whisper', new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
 
-        $formatter->setStyle('whisper', new OutputFormatterStyle($this->grayExists() ? 'gray' : 'blue'));
         $formatter->setStyle('aside', new OutputFormatterStyle('blue'));
         $formatter->setStyle('strong', new OutputFormatterStyle(null, null, ['bold']));
         $formatter->setStyle('return', new OutputFormatterStyle('cyan'));
--- a/src/Output/ShellOutput.php
+++ b/src/Output/ShellOutput.php
@@ -229,7 +229,7 @@ class ShellOutput extends ConsoleOutput
     {
         try {
             $this->write('<fg=gray></>');
-        } catch (InvalidArgumentException $e) {
+        } catch (\InvalidArgumentException $e) {
             return false;
         }
 
--- a/src/Output/ShellOutput.php
+++ b/src/Output/ShellOutput.php
@@ -178,14 +178,14 @@ class ShellOutput extends ConsoleOutput
         $errorFormatter = $this->getErrorOutput()->getFormatter();
         $grayExists = $this->grayExists();
 
-        $formatter->setStyle('info', new OutputFormatterStyle('white', 'blue', ['bold']));
-        $errorFormatter->setStyle('info', new OutputFormatterStyle('white', 'blue', ['bold']));
-        $formatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
-        $errorFormatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
-        $formatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
-        $errorFormatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
-        $formatter->setStyle('whisper', new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
-        $errorFormatter->setStyle('whisper', new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
+        $formatter->setStyle('info', $outputFormatter = new OutputFormatterStyle('white', 'blue', ['bold']));
+        $errorFormatter->setStyle('info', $outputFormatter);
+        $formatter->setStyle('warning', $outputFormatter = new OutputFormatterStyle('black', 'yellow'));
+        $errorFormatter->setStyle('warning', $outputFormatter);
+        $formatter->setStyle('error', $outputFormatter = new OutputFormatterStyle('white', 'red', ['bold']));
+        $errorFormatter->setStyle('error', $outputFormatter);
+        $formatter->setStyle('whisper', $outputFormatter = new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
+        $errorFormatter->setStyle('whisper', $outputFormatter);
 
         $formatter->setStyle('aside', new OutputFormatterStyle('blue'));
         $formatter->setStyle('strong', new OutputFormatterStyle(null, null, ['bold']));
--- a/src/Output/ShellOutput.php
+++ b/src/Output/ShellOutput.php
@@ -181,9 +181,8 @@ class ShellOutput extends ConsoleOutput
         $errorFormatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
         $formatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
         $errorFormatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
-        $formatter->setStyle('whisper', new OutputFormatterStyle($this->grayExists() ? 'gray' : null));
-        $errorFormatter->setStyle('whisper', new OutputFormatterStyle($this->grayExists() ? 'gray' : null));
 
+        $formatter->setStyle('whisper', new OutputFormatterStyle($this->grayExists() ? 'gray' : 'blue'));
         $formatter->setStyle('aside', new OutputFormatterStyle('blue'));
         $formatter->setStyle('strong', new OutputFormatterStyle(null, null, ['bold']));
         $formatter->setStyle('return', new OutputFormatterStyle('cyan'));
--- a/src/Output/ShellOutput.php
+++ b/src/Output/ShellOutput.php
@@ -181,6 +181,8 @@ class ShellOutput extends ConsoleOutput
         $errorFormatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
         $formatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
         $errorFormatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
+        $formatter->setStyle('whisper', new OutputFormatterStyle($this->grayExists() ? 'gray' : null));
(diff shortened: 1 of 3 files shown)


src/Shell.php
```diff
@@ -1201,13 +1223,14 @@ public function getLastExecSuccess(): bool
...
     {
-        $message = $e->getMessage();
-        if (!$e instanceof PsyException) {
-            if ($message === '') {
-                $message = \get_class($e);
-            } else {
-                $message = \sprintf('%s with message \'%s\'', \get_class($e), $message);
-            }
+        if ($e instanceof BreakException) {
+            return sprintf("  <fg=white;bg=blue;options=bold> INFO </> %s.", $e->getRawMessage());
```
bobthecow, line 1227: i'd prefer to use named styles rather than inline styles… for all the reasons classes are better than inline styles in HTML :P

src/Shell.php
```diff
@@ -1201,13 +1223,14 @@ public function getLastExecSuccess(): bool
...
+        } elseif ($e instanceof PsyException) {
+            $message = \sprintf('%s in %s on line %d', $e->getRawMessage(), $e->getFile(), $e->getLine());
```
bobthecow, line 1229: we can probably often (always?) leave off `on line 1` for "in eval()d code" exceptions.

ideally we'd only do that if it was single line input, and couldn't possibly be on any other line, but if that's too hard to get at it's probably fine to just always leave it off for line === 1.

src/Shell.php
```diff
@@ -1253,6 +1287,54 @@ protected function getSeverity(\ErrorException $e): string
...
+                    case \E_DEPRECATED:
+                        return 'Deprecated';
+                    case \E_STRICT:
+                        return 'Strict';
+                }
+            }
+        }
+
+        if ($e instanceof PsyException) {
+            $message = $e->getRawMessage();
```
bobthecow, line 1327: 👆 unused here.

src/Shell.php
```diff
@@ -1253,6 +1287,54 @@ protected function getSeverity(\ErrorException $e): string
...
+            $exceptionShortName = (new \ReflectionClass($e))->getShortName();
+            $typeParts = preg_split('/(?=[A-Z])/', $exceptionShortName);
+            array_pop($typeParts); // Removes "Exception"
```
bobthecow, line 1330: thanks for the comment :)

src/Shell.php
```diff
@@ -1134,6 +1139,22 @@ public function writeReturnValue($ret, bool $rawOutput = false)
...
 
+    /**
+     * Checks if the "gray" color exists on the output.
+     *
+     * @return bool
+     */
+    private function grayExists(): bool
+    {
+        try {
+            $this->output->write('<fg=gray></>');
```
bobthecow, line 1150: probably don't need to try this every time we write something to output :) 

if we're using named styles, maybe we check once when defining the styles (`<whisper>`?) and make it a no-op if grey doesn't work?

src/Shell.php
```diff
@@ -51,10 +52,10 @@ class Shell extends Application
 {
     const VERSION = 'v0.11.7';
 
-    const PROMPT = '>>> ';
-    const BUFF_PROMPT = '... ';
-    const REPLAY = '--> ';
-    const RETVAL = '=> ';
+    const PROMPT = '> ';
+    const BUFF_PROMPT = '. ';
+    const REPLAY = '- ';
```
bobthecow, line 57: The `REPLAY` prompt can get the grey treatment too.

src/Shell.php
```diff
@@ -1435,7 +1517,7 @@ protected function readline(bool $interactive = true)
      */
     protected function getHeader(): string
     {
-        return \sprintf('<aside>%s by Justin Hileman</aside>', $this->getVersion());
+        return \sprintf('<fg=cyan;options=bold>%s by Justin Hileman</>', $this->getVersion());
```
bobthecow, line 1520: this originally was grey, as it wasn't meant to be in your face. but it turns out that's not portable so i changed it to dark blue. No strong opinion on what it should be ¯\\\_(ツ)\_/¯

test/ShellTest.php
```diff
@@ -261,22 +264,25 @@ public function testReportsErrors($errno, $label)
         \rewind($stream);
         $streamContents = \stream_get_contents($stream);
 
-        $this->assertStringContainsString($label, $streamContents);
-        $this->assertStringContainsString('wheee', $streamContents);
-        $this->assertStringContainsString('line 13', $streamContents);
+        $this->assertSame(<<<EOF
```
bobthecow, line 267: these were asserting "contains" because older dependencies don't have the exact same output (see 👇)

bobthecow (commented):
Thanks for the PR!

I think the whitespace (blackspace?) works really well, in general, but is probably annoying on smaller terminal windows, like when running a small shell at the bottom of your IDE. Maybe this can be solved with a `compact` flag in config / CLI options? Or maybe detect the terminal size (when possible) and auto-toggle?

Additionally, the brevity of the single-character prompts is clean, but might make it harder to scan and understand, possibly getting lost in the noise of the shell.

In the past, we've also had requests for customizing colors (usually from people who have issues with the default colors in their particular terminal app).

Between these three, I'm wondering if introducing a concept of themes would be a good idea? Obviously not a blocker for this PR, but might be a way to make it so PsySH can fit with the aesthetic of artisan CLI, but not _have to_? 

This is another reason I like named styles. If someone wants to override what `<info>` is that's a lot easier than hardcoded colors inline.

src/Shell.php
bobthecow: The `⏎` on the end of stdout got the grey/blue treatment as well. If we've got the option for sometimes-grey, we should apply it to there as well.

src/Shell.php
Wirone: Besides named style, 2 little details from me:
```suggestion
            return sprintf('  <fg=white;bg=blue;options=bold> INFO </> %s.', rtrim($e->getRawMessage(), '.'));
```

* Use single quotes, since interpolation is not required
* Ensure there's single `.` at the end (or just remove dot from the message)

Wirone (commented):
Good to see @nunomaduro that you came back with the changes directly to the upstream 👍 More people will take advantage from your improvements 🍻

src/Shell.php
nunomaduro: I've leaved off when line === 1.

test/ShellTest.php
nunomaduro: Can you elaborate on this feedback please?

bobthecow: Check the "minimum deps" versions of the test suite. Output has change enough over the supported range for our dependencies that the exact string for this output doesn't pass for all of them.

nunomaduro: Fixed.

nunomaduro:
Hi @bobthecow, so, here is the list of applied changes, after your first feedback in the pull request:

- The `on line 1` is only displayed when `line > 1`:

Screenshot 2022-08-01 at 08 59 19

- Named styles - the `blue` info background and the `gray` thing - got converted to "named" styles: `info` and `whisper`. Note that, the `whisper` only actually uses `gray` if `gray` is available.

- The `⏎` symbol now uses the `whisper` style. ( can you test this one? ).

- The credits are displayed on `whisper`:
Screenshot 2022-08-01 at 09 11 52

src/Output/ShellOutput.php
```diff
@@ -176,11 +176,16 @@ private function initFormatters()
...
+        $grayExists = $this->grayExists();
 
+        $formatter->setStyle('info', new OutputFormatterStyle('white', 'blue', ['bold']));
+        $errorFormatter->setStyle('info', new OutputFormatterStyle('white', 'blue', ['bold']));
         $formatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
         $errorFormatter->setStyle('warning', new OutputFormatterStyle('black', 'yellow'));
         $formatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
         $errorFormatter->setStyle('error', new OutputFormatterStyle('white', 'red', ['bold']));
+        $formatter->setStyle('whisper', new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
+        $errorFormatter->setStyle('whisper', new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
```
Wirone, line 188: PoC:
```suggestion
        $formatter->setStyle('info', $outputFormatter = new OutputFormatterStyle('white', 'blue', ['bold']));
        $errorFormatter->setStyle('info', $outputFormatter);
        $formatter->setStyle('warning', $outputFormatter = new OutputFormatterStyle('black', 'yellow'));
        $errorFormatter->setStyle('warning', $outputFormatter);
        $formatter->setStyle('error', $outputFormatter = new OutputFormatterStyle('white', 'red', ['bold']));
        $errorFormatter->setStyle('error', $outputFormatter);
        $formatter->setStyle('whisper', $outputFormatter = new OutputFormatterStyle($grayExists ? 'gray' : 'blue'));
        $errorFormatter->setStyle('whisper', $outputFormatter);
```

I'm not happy with consecutive `$outputFormatter` usage, but the idea here is to reuse style between `$formatter` and `$errorFormatter`. Right now it's possible to modify style in one formatter and leave it untouched in the other one, which may lead to inconsistency. It could be improved IMHO.

src/Shell.php
```diff
@@ -1224,7 +1205,7 @@ public function getLastExecSuccess(): bool
     public function formatException(\Exception $e): string
     {
         if ($e instanceof BreakException) {
-            return \sprintf('  <fg=white;bg=blue;options=bold> INFO </> %s.', \rtrim($e->getRawMessage(), '.'));
+            return \sprintf('  <info> INFO </info> %s.', \rtrim($e->getRawMessage(), '.'));
```
Wirone, line 1208: Is it possible to ensure padding inside `<info>` in the style definition, so whitespaces won't be required explicitly here? I mean `<info>INFO</info>`. Or maybe it works already and those space are superfluous here?

Wirone (commented):
Some optional polishing - this is only nice to have 🙂

src/Shell.php
nunomaduro: I am not sure how to accomplish this.

Wirone: Me neither 😅 Probably [`SymfonyStyle`](https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Console/Style/SymfonyStyle.php) could help, I'm pretty sure there is something similar in their commands' output. But it's up to you - from my PoV it can stay like this, without changes 🙂

bobthecow: I don't think so. IIRC it's just for colors.

bobthecow:
> The ⏎ symbol now uses the whisper style. ( can you test this one? ).

This happens when stdout doesn't end with a newline, so you can test with `echo "foo"`.

nunomaduro:
@bobthecow Ready for your second review. ✅ 

Do you know why a few jobs are failing?

Wirone:
@nunomaduro it's caused by indentation of closing HEREDOC (`EOF`). Relaxed syntax was introduced in 7.3 and that's why those versions are green in Actions 🙂

test/ShellTest.php
```diff
@@ -516,16 +519,27 @@ public function testWriteExceptionVerboseButNotReallyBecauseItIsABreakException(
...
-        $shell->writeException(new BreakException('yeah.'));
+        $shell->writeException(new BreakException('yeah'));
         \rewind($stream);
-        $this->assertSame('Exit:  yeah.'.\PHP_EOL, \stream_get_contents($stream));
+        $this->assertSame(<<<EOF
+
+   INFO  yeah.
+
+
+EOF, \stream_get_contents($stream));
```
bobthecow, line 529: in older PHP, you had to end the heredoc with a `;`. So this would need to be something like

```suggestion
        $expected = <<<EOF

   INFO  yeah.


EOF;
        $this->assertSame($expected, \stream_get_contents($stream));
```

nunomaduro:
@bobthecow Let me know if something else is missing here. ✔️

bobthecow:
Sorry, I've been camping and unable to connect to the internet. I'll take a look soon :)

nunomaduro:
Perfect. No worries, and be sure to enjoy your camping. 🏕

bobthecow:
How do we feel about whispering the prompt as well?

Screen Shot 2022-08-13 at 3 54 06 PM

bobthecow:
Thanks for all your work on this :)

PR #738, opened 2022-07-29
Draft: Introduce PHPStan

Wirone:
- static analysis on level 1 (work in progress)
- already existing errors dumped to baseline
- local usage with `make sa`
- GitHub Actions (based on PHPUnit flow) to avoid new errors

-----

ℹ️ INFO: there are maaaaaaany errors in this codebase, even on level 0. Initially I took approach with highest level and baseline file, which allows to avoid new errors without need to fix all existing errors. It also showed scale of a challenge.

There were some errors that couldn't be dumped to baseline and needed to be fixed, but I did not do it - instead those files were excluded from analysis just to be able to go with green PHPStan result.

To bump analysis level:

- bump analysis level in `phpstan.neon.dist`:
- run `make sa`
- fix what's fixable, remaining error can be dumped to baseline by running `vendor/bin/phpstan analyse --generate-baseline=vendor-bin/phpstan/baseline.neon`

PS. you can create `phpstan.neon` with the content:

```neon
includes:
  - phpstan.neon.dist

parameters:
  editorUrl: 'phpstorm://open?file=%%file%%&line=%%line%%'
```

so CLI result will have clickable links that would open file in IDE. Other options (like `level`) also can be overridden there.


bobthecow:
I'm heading out camping for a week, so I won't have a chance to look at this for a bit, but judging from the baseline many of the "errors" are likely due to the minimum PHP version we support. This is only really viable if it can respect that constraint (I'm not sure whether it can or not; I haven't looked into it much).

Wirone:
I'll look at it later 🙂 But on the other hand, is it required to support such old PHP versions? 7.4 is EOL in 4 months... As long ag as you're in `0.*` initial development it's easy to decide since it's not BC-break 😉 Bumping minimal version to 7.4 would allow usage of some language syntax, Rector could help with automated refactoring... What do you think?

Wirone:
@bobthecow there [`phpversion`](https://phpstan.org/config-reference#phpversion) config parameter in PHPStan so it can be enforced which version should be used for finding errors, but unfortunately minimal required version is 7.1:

image

So [question](#issuecomment-1200163579) about minimum required version is even more important 😉

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,66 @@
...
+        default: false
+
+jobs:
+  test:
+    name: PHPStan ${{ matrix.php }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        php: ['7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1']
```
keradus, line 22: is it really needed to run on each PHP version? or maybe enough if run for lowest / highest only?

Makefile
```diff
@@ -35,6 +35,9 @@ test: ## Run unit tests
 test: vendor/bin/phpunit
 	$< $(PHPUNIT_OPTS)
 
+sa: ## Run Static Analysis
```
keradus, line 38: fun fact - I always saw it as Static Code Analysis. It took me a while to figure out what `sa` shortcut stands for

.github/workflows/phpstan.yml
Wirone: This is good question. Personally I think it's required to run it on every PHP version that is supported because it will detect:
- usage of PHP features not available in specific versions. Of course, this can be also detected by unit tests, but we can't be sure whole codebase is covered with tests, while SA will look at every file.
- inconsistency between vendor packages' signatures: this is most problematic IMHO, for each PHP version set of dependencies is different because of wide constraints. And SA should detect if this code is compatible with every vendor version in supported range... But it is not that easy, [look at this thread](https://mobile.twitter.com/_Codito_/status/1553753386985705484) 🤔

As far as I remember this approach is used in several projects, I can look for examples if you want 😉 Alternative is to write in 8.1 and then use Rector for downgrading to 7.0 (like PHPStan and Rector do), but it is A LOT of work.

Makefile
Wirone: Yeah, but you don't do Code Unit Tests, right 😝? Personally I always use just "static analysis" term, code is implicit there. But yeah, it's subjective 🙂

Wirone:
FYI: I have some fixes for level 0, but [PHPStan's approach](https://twitter.com/_Codito_/status/1553753395160309761) makes it harder to clear this level. I need to think how to solve it, but most probably some files need to be excluded from the analysis. I will commit it later this evening probably 🙂

ondrejmirtes:
I wouldn't really go all-in on the highest level with the help of the baseline if the maintainers and contributors never used static analysis on this project before. It's better to dip your toes in and go dutifully from the lowest level 0, learning about the advantages of static analysis gradually, and fix useful bugs, rather than suddenly having all your code checked and getting weird errors in subsequent PRs that are not really caused by the authors of those PRs.

The current approach is this PR is prone to create more headaches rather than solving useful problems.

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
+includes:
+  - vendor-bin/phpstan/baseline.neon
+
+parameters:
+  # Enforce minimal supported version (note: 7.0 can't be set in PHPStan)
+  phpVersion: 70100
```
ondrejmirtes, line 6: There's no point in running PHPStan multiple times on multiple PHP versions if 7.1 is forced in the configuration like this.

ondrejmirtes:
I'm the author and main maintainer of PHPStan. I really don't like how PHPStan is set up here.

Compare this PR that ignores several thousands errors without giving the contributors any hints how to continue and what to do if these errors re-appear in future work, with the output you get if you simply run level 0 just on `src`: https://gist.github.com/ondrejmirtes/e8fdeadf0b54d67551457c2ea5f4be71

You get 31 errors that are obvious and simple (and beneficial!) to fix. I'd start there - introduce PHPStan to CI on level 0, fix those 31 errors, merge the PR. In future PRs you can:

* Introduce analysis to tests too. It would require you to install phpstan-phpunit extension as well.
* Increase level to 1, then 2, etc...
* If you really feel like it, you can start improving your code with phpstan-strict-rules, but it isn't necessary.

Wirone:
Thanks @ondrejmirtes for your valuable feedback 👍 As you can see, it's only Draft 😉 I wanted to discuss overall will for introducing SA and of course the approach. Also, [as I wrote earlier](#issuecomment-1201083491), I've fixed several issues on level 0 but got stuck because of how PHPStan handles "fatal" errors. So basically baselines are here for now just to show scale of the challenge 😅 In the end I wanted to deliver fixes for reasonable level (about 6) with some potential more demanding errors dumped to the baseline, so they could be fixed later.

PS. Thanks for pointing `phpstan/phpstan-phpunit`, I forgot to add it, so I'll do it later 🙂

ondrejmirtes:
When I run PHPStan myself, I get these 6 errors that cannot be put into the baseline:

``` ------ -------------------------------------------------------------------------------------------
  Line   CodeCleaner/ValidClassNamePass.php
 ------ -------------------------------------------------------------------------------------------
  272    Method Psy\CodeCleaner\ValidClassNamePass::getScopeType() should return string but return
         statement is missing.
 ------ -------------------------------------------------------------------------------------------

 ------ --------------------------------------------------------------------------------------------
  Line   Configuration.php
 ------ --------------------------------------------------------------------------------------------
  1166   Method Psy\Configuration::getInputInteractive() should return bool but return statement is
         missing.
 ------ --------------------------------------------------------------------------------------------

 ------ ----------------------------------------------------------------------------------------------
  Line   Readline/Hoa/FileFinder.php
 ------ ----------------------------------------------------------------------------------------------
  520    Return type mixed of method Psy\Readline\Hoa\FileFinder::getIterator() is not covariant with
         tentative return type Traversable of method IteratorAggregate::getIterator().
         💡 Make it covariant, or use the #[\ReturnTypeWillChange] attribute to temporarily suppress
         the error.
 ------ ----------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------
  Line   Readline/Hoa/FileLink.php
 ------ -------------------------------------------------------------------------------------
  141    Parameter #2 $target of method Psy\Readline\Hoa\FileLink::create() is not optional.
 ------ -------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------
  Line   Util/Str.php
 ------ -------------------------------------------------------------------------------------------
  105    Method Psy\Util\Str::unvisReplace() should return string but return statement is missing.
 ------ -------------------------------------------------------------------------------------------

 ------ ---------------------------------------------------------------------------------------------
  Line   VersionUpdater/IntervalChecker.php
 ------ ---------------------------------------------------------------------------------------------
  48     Method Psy\VersionUpdater\IntervalChecker::getDateInterval() should return DateInterval but
         return statement is missing.
 ------ ---------------------------------------------------------------------------------------------
```

Which one did you have a problem fixing?

phpstan.neon.dist
Wirone: Thanks for the tip, I didn't know it. I just wanted to point PHP version since it can't be automatically determined from `composer.json` because PHPStan is not installed as project's dev dependency.

Wirone:
I have 9 for `src` (running on 7.4 currently, but I experiment with different versions):

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 
  Line   Readline/Hoa/FileGeneric.php                                                                                                                                                  
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 
  232    Parameter #1 $time (int|null) of method Psy\Readline\Hoa\FileGeneric::touch() does not match parameter #1 $time (int) of method Psy\Readline\Hoa\StreamTouchable::touch().    
  232    Parameter #2 $atime (int|null) of method Psy\Readline\Hoa\FileGeneric::touch() does not match parameter #2 $atime (int) of method Psy\Readline\Hoa\StreamTouchable::touch().  
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 

 ------ ------------------------------------------------------------------------------------- 
  Line   Readline/Hoa/FileLink.php                                                            
 ------ ------------------------------------------------------------------------------------- 
  141    Parameter #2 $target of method Psy\Readline\Hoa\FileLink::create() is not optional.  
 ------ ------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------- 
  Line   Reflection/ReflectionClassConstant.php                                                           
 ------ ------------------------------------------------------------------------------------------------- 
  57     Parameter #1 $class of method Psy\Reflection\ReflectionClassConstant::export() is not optional.  
  57     Parameter #2 $name of method Psy\Reflection\ReflectionClassConstant::export() is not optional.   
 ------ ------------------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------------- 
  Line   Reflection/ReflectionConstant_.php                                                          
 ------ -------------------------------------------------------------------------------------------- 
  66     Parameter #1 $name of method Psy\Reflection\ReflectionConstant_::export() is not optional.  
 ------ -------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------------- 
  Line   Reflection/ReflectionLanguageConstruct.php                                                          
 ------ ---------------------------------------------------------------------------------------------------- 
  91     Parameter #1 $name of method Psy\Reflection\ReflectionLanguageConstruct::export() is not optional.  
 ------ ---------------------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------------- 
  Line   Reflection/ReflectionNamespace.php                                                          
 ------ -------------------------------------------------------------------------------------------- 
  46     Parameter #1 $name of method Psy\Reflection\ReflectionNamespace::export() is not optional.  
 ------ -------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------- 
  Line   Util/Str.php                                                                               
 ------ ------------------------------------------------------------------------------------------- 
  105    Method Psy\Util\Str::unvisReplace() should return string but return statement is missing.  
 ------ ------------------------------------------------------------------------------------------- 
```

All those issues are related to signatures and I did not want to touch them since I don't know this codebase.

image
image

From my perspective it would be better if @bobthecow could resolve those problems 😉 Hence lack of possibility to add these to baseline is blocking further works (or I will just exclude those files from analysis just to go to next level)

.github/workflows/phpstan.yml
Wirone: But @ondrejmirtes [says](#discussion_r935470197) it's not required 😉 Unfortunately, PHP 7.1 can be enforced by config and this library supports 7.0 🤔

ondrejmirtes:
Nice thing about level 0 is that you don't need to care about them now...

Wirone:
@ondrejmirtes I don't understand 🤔 I get those errors when running on level 0 without baseline (and with my not committed fixes for the remaining errors). So yeah, it's something I need to care about now 😅 But as I said, I'll exclude those files because I don't want to deal with this. Will push some fixes later.

Wirone:
@bobthecow OK, I'll pause here and wait for your review 🙂 There are some thoughts in commits' messages, especially about `Readline` interface (I thought enforcing constructor's signature would fix errors, but [it didn't](https://phpstan.org/r/f76c8dbe-5b1c-451d-8d37-a5c6a86edd1a)). Maybe we can bring up solutions to errors dump into baseline file and/or discuss fixes already applied.

phpstan.neon.dist
Wirone: @ondrejmirtes I've just tested PHPStan on current branch's state (green level 1) but with other PHP versions (8.0.20 and 8.1.7) and I got 23 errors, internal errors included (`Internal error: Internal error: syntax error, unexpected token "match" in file test/Util/DocblockTest.php` - wtf, there isn't even `match` in this file 😅). So it looks like there is difference between PHP/deps versions.

@bobthecow there's also problem that PHPStan does not allow to configure 7.0 here and also [`phpstan/phpstan-phpunit`](https://packagist.org/packages/phpstan/phpstan-phpunit) requires 7.2... So it really would help if PsySH supported only 7.4+ 🙂

ondrejmirtes: Just run SA once on for example 8.1 (if it's supported) with the right deps and don't bother with anything more complicated.

Wirone:
@bobthecow did you have time to look at it? What do you think?

bobthecow:
Gaahhhhhh where has the summer gone? So sorry about the delay 🙁

src/Command/ShowCommand.php
```diff
@@ -29,14 +29,6 @@ class ShowCommand extends ReflectingCommand
     private $lastException;
     private $lastExceptionIndex;
 
-    /**
-     * @param string|null $colorMode (deprecated and ignored)
-     */
-    public function __construct($colorMode = null)
```
bobthecow, line 35: This is intentionally preventing `$colorMode` from being passed to `Symfony\Component\Console\Command\Command`, which expects an optional `$name` instead :-/

src/Configuration.php
```diff
@@ -1154,6 +1154,8 @@ public function getOutputDecorated()
             case self::COLOR_MODE_DISABLED:
                 return false;
         }
+
+        return null;
```
bobthecow, line 1158: `colorMode()` always returns a value, and it's always one of the above; the switch narrows the possible types down to "never". i get that this might not be the easiest thing to statically analyze, but in the event that the return type for `colorMode` changed, this would give the wrong answer anyway :) 

I think the right fix is to move the `COLOR_MODE_AUTO` case to the bottom and add a `default`. Is that sufficient to get PHPStan to stop complaining?

src/ConsoleColorFactory.php
```diff
@@ -1,39 +0,0 @@
...
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-namespace Psy;
-
-use JakubOnderka\PhpConsoleColor\ConsoleColor;
-
-/**
- * @deprecated Nothing should use this anymore
```
bobthecow, line 17: this has to stick around until the next version bump :)

src/Readline/Hoa/ConsoleWindow.php
```diff
@@ -120,7 +120,7 @@ public static function getSize(): array
         }
 
         $command = $term.'tput cols && '.$term.'tput lines';
-        $tput = Processus::execute($command, false);
+        $tput = ConsoleProcessus::execute($command, false);
```
bobthecow, line 123: 🎉 an actual bug :P

src/Readline/Hoa/FileFinder.php
```diff
@@ -46,7 +46,7 @@ class FileFinder implements \IteratorAggregate
     /**
      * SplFileInfo classname.
      */
-    protected $_splFileInfo = SplFileInfo::class;
+    protected $_splFileInfo = IteratorSplFileInfo::class;
```
bobthecow, line 49: should this be `\SplFileInfo`?

src/Readline/Hoa/Xcallable.php
```diff
@@ -117,10 +117,10 @@ public function __construct($call, $able = '')
                 }
             } elseif (\is_array($call) && isset($call[0])) {
                 if (!isset($call[1])) {
-                    return $this->__construct($call[0]);
+                    $this->__construct($call[0]);
```
bobthecow, line 120: the point of the return here (and below) was to prevent execution from continuing. if we're not going to return the result of `__construct` (which is harmless, but ¯\\\_(ツ)\_/¯) then we need a bare `return` on the next line.

vendor-bin/phpstan/baseline.neon
```diff
@@ -0,0 +1,56 @@
...
+	ignoreErrors:
+		-
+			message: "#^Variable \\$_ might not be defined\\.$#"
+			count: 1
+			path: ../../src/Command/TimeitCommand.php
+
+		-
+			message: "#^Variable \\$_style might not be defined\\.$#"
+			count: 2
+			path: ../../src/Readline/Hoa/ConsoleCursor.php
```
bobthecow, line 11: let's fix this one by adding a `default` to "block" case in the switch.

bobthecow (commented):
In general, this is fine. I'd defer to @ondrejmirtes on how to best run this.

src/Configuration.php
Wirone: Yes, it would be enough 👍

src/Readline/Hoa/FileFinder.php
Wirone: Dunno 🤷 I'll look at it.

bobthecow:
This has been merged, with the suggested changes.

I ended up going with a single test run on PHP 7.4 for simplicity. That will likely be the minimum target for the next major release, and messing with a lower PHP version right now probably isn't worth it 🙂

Thanks so much for your work @Wirone!

Wirone:
Happy I could help and even find some issues 🙂 Are you interested in bumping analysis level in the future?

bobthecow:
Possibly! Open a PR and we can discuss 😉

bobthecow:
I did play with higher levels, and it got silly pretty quickly due to the lack of `void` return types in our minimum supported PHP version.

PR #740, opened 2022-08-05
Do not spawn a shell to find the 'less' executable

### LINKED ISSUE #736: Don't call fork+execve just to find a binary
Right now, there is this code:
https://github.com/bobthecow/psysh/blob/f455acf3645262ae389b10e9beba0c358aa6994e/src/Configuration.php#L1219

There, `which` is invoked from a subshell, just to get the full path of `less` (in fact 2 fork+execve's).

In a couple of lines of php, the functionality of which can be reproduced, and it's ridiculously much faster.

Compare the following two:
```php
<?php // whichexec.php 
for ($i = 0; $i < 5000; ++$i)
    print exec('which less') . "\n";
```
```php
<?php // which.php 
function which($program)
{
    $path = getenv('PATH');
    if (empty($path))
        $path = '/usr/sbin:/usr/bin:/sbin:/bin';
    $paths = explode(':', $path);
    foreach ($paths as $path) {
        $fullpath = "$path/$program";
        $st = @stat($fullpath);
        if ($st != null &&
                ($st['mode'] & 0770000) == 0100000 &&
                ($st['mode'] & 0555) != 0) {
            return $fullpath;
        }
    }
    return null;
}
for ($i = 0; $i < 5000; ++$i)
    print which('less') . "\n";
```
And these runs:
```console
$ time php whichexec.php  >/dev/null 

real	0m11.103s
user	0m8.958s
sys	0m2.705s
```
```console
$ time php which.php  >/dev/null 

real	0m0.035s
user	0m0.023s
sys	0m0.012s
```

### Why do I care?

Yes. You do not run `which` in a 5000x loop. But that's still not a reason to do slow stuff when a faster/easy alternative is available.

_But, more importantly:_

When running auditing tools like [Falco](https://github.com/falcosecurity/falco/) it trips over unnecessary calls to binaries like this.

Therefore I ask you to be nice to the auditors of tools and _only_ make calls to binaries when absolutely necessary.

For instance by doing this:
```diff
--- Configuration.php.orig	2022-07-29 14:15:42.072144525 +0200
+++ Configuration.php	2022-07-29 14:17:00.799744451 +0200
@@ -1216,7 +1216,7 @@ class Configuration
             if ($pager = \ini_get('cli.pager')) {
                 // use the default pager
                 $this->pager = $pager;
-            } elseif ($less = \exec('which less 2>/dev/null')) {
+            } elseif ($less = self::which('less')) {
                 // check for the presence of less...
                 $this->pager = $less.' -R -F -X';
             }
@@ -1823,4 +1823,22 @@ class Configuration
 
         return $mode === 0010000 || $mode === 0040000 || $mode === 0100000 || $mode === 0120000;
     }
+
+    private static function which($program)
+    {
+        $path = getenv('PATH');
+        if (empty($path))
+            $path = '/usr/sbin:/usr/bin:/sbin:/bin';
+        $paths = explode(':', $path);
+        foreach ($paths as $path) {
+            $fullpath = "$path/$program";
+            $st = @stat($fullpath);
+            if ($st != null &&
+                    ($st['mode'] & 0770000) == 0100000 &&
+                    ($st['mode'] & 0555) != 0) {
+                return $fullpath;
+            }
+        }
+        return null;
+    }
 }
```

See a similar issue at https://github.com/symfony/symfony/issues/47109 . Thanks for making the world a better place! :beers: 

Walter Doekes
OSSO B.V.

wdoekes:
Look it up ourselves instead.

Closes #736.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -1216,7 +1216,7 @@ class Configuration
             if ($pager = \ini_get('cli.pager')) {
                 // use the default pager
                 $this->pager = $pager;
-            } elseif ($less = \exec('which less 2>/dev/null')) {
+            } elseif ($less = self::which('less')) {
                 // check for the presence of less...
                 $this->pager = $less.' -R -F -X';
             }
@@ -1822,5 +1822,32 @@ class Configuration
         $mode = $stat['mode'] & 0170000;
 
         return $mode === 0010000 || $mode === 0040000 || $mode === 0100000 || $mode === 0120000;
+    }
+
+    /**
+     * Locate a command (an executable) in $PATH.
+     *
+     * Behaves like 'command -v COMMAND' or 'which COMMAND'.
+     * If $PATH is unset/empty it defaults to '/usr/sbin:/usr/bin:/sbin:/bin'.
+     *
+     * @param string $command the executable to locate
+     *
+     * @return string
+     */
+    private static function which($command)
+    {
+        $path = \getenv('PATH');
+        if (empty($path)) {
+            $path = '/usr/sbin:/usr/bin:/sbin:/bin';
+        }
+        $paths = \explode(\PATH_SEPARATOR, $path);
+        foreach ($paths as $path) {
+            $fullpath = $path.\DIRECTORY_SEPARATOR.$command;
+            if (@\is_file($fullpath) && @\is_executable($fullpath)) {
+                return $fullpath;
+            }
+        }
+
+        return null;
     }
 }


bobthecow:
Great, thanks! Do you mind adding a test case for this?

wdoekes:
> Do you mind adding a test case for this?

Actually I do. I have no idea how your test system works and I generally don't use php. (I don't use your software myself.)

I figure you can create a test case in 5 minutes, whereas it will take me at least an hour to figure out how to run/build tests here :)

bobthecow:
Fair enough :)

src/Configuration.php
```diff
@@ -1823,4 +1823,34 @@ private static function looksLikeAPipe($stream): bool
...
+        $path = \getenv('PATH');
+        if (empty($path)) {
+            $path = '/usr/sbin:/usr/bin:/sbin:/bin';
+        }
+        $paths = \explode(':', $path);
+        foreach ($paths as $path) {
+            $fullpath = "$path/$command";
+            $st = @\stat($fullpath);
+            if ($st !== false &&
+                    ($st['mode'] & 0170000) === 0100000 &&   // S_IFREG
```
bobthecow, line 1848: should we also include symlinks?

src/Configuration.php
```diff
@@ -1823,4 +1823,34 @@ private static function looksLikeAPipe($stream): bool
...
+                    ($st['mode'] & 0555) !== 0) {            // any executable bit
```
bobthecow, line 1849: i threw together a simple test fixture and it `$st['mode'] & 0555` seems to be true for regular files even with no executable bit set. i do get the expected result for `is_executable`.

it looks like `is_file`, `is_link` and `is_executable` uses `stat()` under the hood (and cache the result) so should we use those directly instead? 

```
if ((\@is_file($fullpath) || \@is_link($fullpath)) && \@is_executable($fullpath)) {
    return $fullpath;
}
```

src/Configuration.php
wdoekes: `stat()` follows symlinks (we didn't use `lstat()`), so you would never encounter a symlink:
```
$ ls -l /bin/sh
lrwxrwxrwx 1 root root 4 mrt 23 14:49 /bin/sh -> dash
```
```
$ python3 -c 'import os; s=os.stat("/bin/sh"); print(oct(s.st_mode))'
0o100755
```
vs.
```
$ python3 -c 'import os; s=os.lstat("/bin/sh"); print(oct(s.st_mode))'
0o120777
```

src/Configuration.php
wdoekes: Ha. I threw myself off there.
`($st['mode'] & 0111) !== 0)` was the original idea, but I figured we need to be able to read the executables too. (And then I forgot that we couldn't do a simple boolean compare anymore.)

But I guess we don't need `04`:
```
$ ls -l /usr/bin/zip
---x--x--x 1 root root 203768 mrt 25 10:55 /usr/bin/zip
```
```
$ which zip
/usr/bin/zip
```
```
$ zip
Copyright (c) 1990-2008 Info-ZIP - Type 'zip "-L"' for software license.
...
```
So that should just be `0111` then. 

But for anything other than `0001` the user/group needs to be checked too: so you'd need an `access(..., X_OK)` call as well.

`which` uses `newfstatat` + `faccessat2(..., X_OK, ...)`.
`is_file` uses `newfstatat` and `is_executable` uses `access(..., X_OK)`, so that is a perfect match.

Fixing...

wdoekes: > But I guess we don't need 04:

Okay, we would, if it is a script. But we don't need to be smarter than `which` here.
```
$ which less
/home/walter/bin/less
```
```
$ less
/bin/sh: 0: cannot open /home/walter/bin/less: Permission denied
```
Ignoring that...

bobthecow:
Thanks!

PR #743, opened 2022-09-15
Add support for disabling the header in interactive sessions

jasongill:
This is a simple patch that allows for a new configuration variable, `showHeader`, which can be set to false to disable the `Psy Shell v0.11.8 (PHP 8.1.10 — cli) by Justin Hileman` header in interactive sessions.

The header is nice, but honestly some times I just want to jump in and out of psysh with the extra row cluttering up the terminal history, so my `showHeader` option pairs nicely with the new compact mode.

Plus, now with `'showHeader'=>false, 'startupMessage'=>"This is my startup message, thanks Justin Hileman for PySH!"` in the config file, you can customize the "header" even more than before.

Thanks for the hard work on this project and thanks for considering my pull request 👍🏻

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -53,6 +53,7 @@ class Configuration
         'compactOutput',
         'dataDir',
         'defaultIncludes',
+        'showHeader',
         'eraseDuplicates',
         'errorLoggingLevel',
         'forceArrayIndexes',
@@ -86,6 +87,7 @@ class Configuration
     /** @var string|false */
     private $historyFile;
     private $historySize;
+    private $showHeader = true;
     private $eraseDuplicates;
     private $manualDbFile;
     private $hasReadline;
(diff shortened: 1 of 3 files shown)


bobthecow:
Maybe I'll make hiding the header a paid feature of PsySH Pro 🤔

bobthecow:
Thanks for your contribution. I've decided I won't be merging it as this header is often the only indication to users that they are using PsySH. See https://github.com/laravel/tinker/pull/152#discussion_r931263140

The header _is_ more subdued as of the next release, thanks to #737.

jasongill:
@bobthecow Fair enough; would you consider making it so that using your existing 'startupMessage' option would replace the header? That would solve all issues - downstream packages like Laravel could still customize the output as they need, and users could still be alerted that they are inside of PsySH (while also allowing power users to modify the header to be a better fit for their environment)

PR #744, opened 2022-10-23
Adds the psysh download url to the new-version-available message

gefd:
Minor QoL text update adds the "psysh.org/psysh" url to the new version available message

### DIFF
--- a/src/Shell.php
+++ b/src/Shell.php
@@ -1609,7 +1609,7 @@ class Shell extends Application
         try {
             $client = $this->config->getChecker();
             if (!$client->isLatest()) {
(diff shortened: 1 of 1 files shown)


src/Shell.php
```diff
@@ -1609,7 +1609,7 @@ protected function writeVersionInfo()
         try {
             $client = $this->config->getChecker();
             if (!$client->isLatest()) {
-                $this->output->writeln(\sprintf('<whisper>New version is available (current: %s, latest: %s)</whisper>', self::VERSION, $client->getLatest()));
+                $this->output->writeln(\sprintf('<whisper>New version is available @ psysh.org/psysh (current: %s, latest: %s)</whisper>', self::VERSION, $client->getLatest()));
```
bobthecow, line 1612: let's spell out `at` rather than using `@` (since it's only one more character :P)

bobthecow:
Thank you!

PR #745, opened 2022-10-24
Add self-update option to replace the current version with the latest available version

gefd:
This adds a --self-update option that will download the latest version from github and replace the currently installed version.

The default downloader is a very simple curl implementation, with a fallback to file_get/put_contents if curl is not installed, and 
will throw an exception if allow_url_fopen is not true. The download process will save the downloaded file to the system temp
directory and extract the archive contents to the location of the current installed version.

The self-update option includes a --keep-backup that will keep the temporary backup copy of the current version in place after
updating.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -19,6 +19,9 @@ use Psy\Output\ShellOutput;
 use Psy\TabCompletion\AutoCompleter;
 use Psy\VarDumper\Presenter;
 use Psy\VersionUpdater\Checker;
+use Psy\VersionUpdater\GithubChecker;
+use Psy\VersionUpdater\IntervalChecker;
+use Psy\VersionUpdater\NoopChecker;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -1562,7 +1565,7 @@ class Configuration
             $interval = $this->getUpdateCheck();
             switch ($interval) {
                 case Checker::ALWAYS:
-                    $this->checker = new Checker\GitHubChecker();
+                    $this->checker = new GitHubChecker();
                     break;
 
                 case Checker::DAILY:
@@ -1570,14 +1573,14 @@ class Configuration
                 case Checker::MONTHLY:
                     $checkFile = $this->getUpdateCheckCacheFile();
                     if ($checkFile === false) {
-                        $this->checker = new Checker\NoopChecker();
+                        $this->checker = new NoopChecker();
                     } else {
-                        $this->checker = new Checker\IntervalChecker($checkFile, $interval);
+                        $this->checker = new IntervalChecker($checkFile, $interval);
                     }
                     break;
 
                 case Checker::NEVER:
-                    $this->checker = new Checker\NoopChecker();
+                    $this->checker = new NoopChecker();
                     break;
             }
         }
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -19,7 +19,7 @@ use Psy\Output\ShellOutput;
 use Psy\TabCompletion\AutoCompleter;
 use Psy\VarDumper\Presenter;
 use Psy\VersionUpdater\Checker;
-use Psy\VersionUpdater\GithubChecker;
+use Psy\VersionUpdater\GitHubChecker;
 use Psy\VersionUpdater\IntervalChecker;
 use Psy\VersionUpdater\NoopChecker;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -19,9 +19,6 @@ use Psy\Output\ShellOutput;
 use Psy\TabCompletion\AutoCompleter;
 use Psy\VarDumper\Presenter;
 use Psy\VersionUpdater\Checker;
-use Psy\VersionUpdater\GitHubChecker;
-use Psy\VersionUpdater\IntervalChecker;
-use Psy\VersionUpdater\NoopChecker;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -1563,7 +1560,7 @@ class Configuration
             $interval = $this->getUpdateCheck();
             switch ($interval) {
                 case Checker::ALWAYS:
-                    $this->checker = new GitHubChecker();
+                    $this->checker = new Checker\GitHubChecker();
                     break;
 
                 case Checker::DAILY:
@@ -1571,14 +1568,14 @@ class Configuration
                 case Checker::MONTHLY:
                     $checkFile = $this->getUpdateCheckCacheFile();
                     if ($checkFile === false) {
-                        $this->checker = new NoopChecker();
+                        $this->checker = new Checker\NoopChecker();
                     } else {
-                        $this->checker = new IntervalChecker($checkFile, $interval);
+                        $this->checker = new Checker\IntervalChecker($checkFile, $interval);
                     }
                     break;
 
                 case Checker::NEVER:
-                    $this->checker = new NoopChecker();
+                    $this->checker = new Checker\NoopChecker();
                     break;
             }
         }
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -358,6 +358,8 @@ class Configuration
             new InputOption('no-interaction', null, InputOption::VALUE_NONE, 'Run PsySH without interactive input. Requires input from stdin.'),
             new InputOption('raw-output', 'r', InputOption::VALUE_NONE, 'Print var_export-style return values (for non-interactive input)'),
 
+            new InputOption('self-update', 'u', InputOption::VALUE_NONE, 'Update to the latest version'),
+
             new InputOption('yolo', null, InputOption::VALUE_NONE, 'Run PsySH with minimal input validation. You probably don\'t want this.'),
         ];
     }
--- a/src/Shell.php
+++ b/src/Shell.php
@@ -130,9 +130,7 @@ class Shell extends Application
      */
     public static function isPhar(): bool
     {
-        $self = \realpath($_SERVER['argv'][0] ?? \getenv('_'));
-
-        return \class_exists("\Phar") && \Phar::running(false) === $self;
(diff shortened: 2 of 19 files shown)


src/VersionUpdater/SelfUpdate.php
```diff
@@ -0,0 +1,134 @@
...
+                \unlink($downloadedFile);
+            }
+        } else {
+            $output->write(" <info>OK</info>\n");
+        }
+
+        $pharArchive = new \PharData($downloadedFile);
+        if (!$pharArchive->valid()) {
+            \unlink($downloadedFile);
+            $output->writeln('<error>Downloaded file is not a valid archive</error>');
```
bobthecow, line 100: let's write output before unlinking, just in case.

bobthecow (changes requested):
Amazing, thank you! A couple of minor tweaks, but one bigger change:

PsySH can be installed ~4 ways: phar, global composer dep, local composer dep, and system packages (we can mostly ignore the last one though). Self-update only really makes sense for the phar-flavored installs. We should ensure that it doesn't run when we're installed another way.

Maybe simple as wrapping the actual SelfUpdate run call with something that ensures `\Phar::running()`? Or better validation to make sure we're running inside _the PsySH phar_?

Additionally, we should avoid showing `--self-update` as an available option in `--help` output if we won't actually self update :)

gefd:
@bobthecow Thanks for the review and feedback. I will provide an update in the next day or two :)

src/Shell.php
```diff
@@ -1595,7 +1603,6 @@ private function addMatchersToAutoCompleter(array $matchers)
     }
 
     /**
-     * @todo Implement self-update
```
bobthecow, line 1598: <3

src/Configuration.php
```diff
@@ -19,7 +19,7 @@
 use Psy\TabCompletion\AutoCompleter;
 use Psy\VarDumper\Presenter;
 use Psy\VersionUpdater\Checker;
-use Psy\VersionUpdater\GitHubChecker;
+use Psy\VersionUpdater\GithubChecker;
```
bobthecow, line 22: `GitHubChecker` (case sensitive filesystems or class loaders might 💥)

src/VersionUpdater/SelfUpdate.php
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * @param InputInterface  $input
+     * @param OutputInterface $output
+     *
+     * @return int
+     */
+    public function run(InputInterface $input, OutputInterface $output)
+    {
+        $oldVersion = Shell::VERSION;
+        $installLocation = \realpath($_SERVER['argv'][0]);
```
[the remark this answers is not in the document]
gefd: I think this is addressed by verifying the currently executing script is a phar archive? I could make the isPhar check more explicit by comparing argv:0 to Phar::running(false), which I think would be accurate in this case?

src/Shell.php
gefd: :+1:

test/VersionUpdater/SelfUpdateTest.php
```diff
@@ -0,0 +1,254 @@
...
+            new InputOption('verbose', 'v', InputOption::VALUE_NONE)
+        ]));
+
+        return $input;
+    }
+
+    private function getMockChecker(bool $isLatest = false, string $version = Shell::VERSION)
+    {
+        $checker = $this->getMockBuilder(Checker::class)
+            ->onlyMethods(['getLatest', 'isLatest'])
```
bobthecow, line 194: `onlyMethods` isn't a thing on PHPUnit versions old enough to run on 7.0 and 7.1. Honestly, I'd just skip these tests on PHP <7.2 if it's a pain to do another way.

src/VersionUpdater/Installer.php
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * @var string
+     */
+    protected $tempDirectory;
+
+    public function __construct(string $tempDirectory = null)
+    {
+        $this->tempDirectory = $tempDirectory ?: \sys_get_temp_dir();
+        $this->installLocation = \realpath($_SERVER['argv'][0]);
```
bobthecow, line 31: should we use `\Phar::running(true)`?

src/VersionUpdater/SelfUpdate.php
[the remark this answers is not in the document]
bobthecow: oh, yeah, this is what i was asking about below. this comment got collapsed :)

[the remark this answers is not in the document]
bobthecow: if we do add that check, the error should be verbose enough for people to pinpoint the issue when argv 0 doesn't match the currently running phar.

src/VersionUpdater/Installer.php
gefd: For sure, but it would have to be `\Phar::running(false)` since fs operations won't work with 'phar://' uri's

bobthecow: sorry, yeah. got the argument flipped.

src/Shell.php
```diff
@@ -125,6 +125,16 @@ public static function isIncluded(array $trace): bool
...
     }
 
+    /**
+     * Check if the currently running PsySH bin is a phar archive.
+     */
+    public static function isPhar(): bool
+    {
+        $self = \realpath($_SERVER['argv'][0] ?? \getenv('_'));
+
+        return \class_exists("\Phar") && \Phar::running(false) === $self;
```
bobthecow, line 135: good call. this feels a lot better.

gefd:
I have updated the Shell::isPhar method to check `\Phar::running(false)` against `\realpath($_SERVER['argv'][0] ?? \getenv('_'))`, though I can think some cases where this may be still true when it shouldn't be. For example; if a different.phar included psysh some how and then convinced the self-update process to run, different.phar would be replaced by psysh. Which, depending on what different.phar is, might be a good thing ;)

I suppose running `exec(\Phar::running(false).' --version')` and comparing that to `Shell::getVersionHeader()` would be the most accurate method of making sure the target is a psysh.phar. This could be overkill though... If some body has managed to execute the case above - I think it's safe to assume they either ment to cause the result, or they deserve weird stuff to happen anyway ;)

bobthecow:
yeah, i can't think of many ways to break this, as it currently stands.  maybe trying to load a known file from `\Phar::running()`…

oh! `strpos(__FILE__, \Phar::running()) === 0` is true inside `\Psy\Shell::isPhar`. We might not need to bring argv into it?

bobthecow:
that approach would make it so some level of indirection (e.g. a bash script that, itself, runs psysh) would be fine, as the `Phar::running()` / `__FILE__` should be the source of truth about what, exactly, we're trying to replace?

gefd:
I had considered using `__FILE__` but thought that would be redundant since it returns the path of the current php file, but _of course_ the path prefix of `__FILE__` is the phar file!

bobthecow (approved):
Things look good to me. Is there anything else you wanted to update before merging?

src/VersionUpdater/Installer.php
```diff
@@ -0,0 +1,143 @@
...
+     * Extract the "psysh" phar from the archive and move it, replacing the currently installed phar.
+     *
+     * @param string $sourceArchive
+     *
+     * @return bool
+     */
+    public function install(string $sourceArchive): bool
+    {
+        $pharArchive = new \PharData($sourceArchive);
+        $pharArchive->extractTo($this->tempDirectory, ['psysh'], true);
```
bobthecow, line 91: we should probably extract this to a `tempnam` directory _inside_ the temp dir, because there just might be a folder called `psysh` there already.

```
PharData::extractTo(/var/folders/lx/7kfh1rt500z6yphfzdm38f6w0000gn/T/psysh): Failed to open stream: Is a directory
```

bobthecow (changes requested):

gefd:
There was one minor change I had considered - when building the file name for the download URL. Currently it will simply try to download the file and, in the case you highlighted - where a "flavoured" version may not have an update available, this will simply cause a 404 error from the downloader and fail the self-update process...

The change I'd considered was updating the GitHubChecker to keep the parsed latest.json it retrieved when checking for the latest version in a private variable and checking the filename built from "latest-version + flavour-postfix + .tar.gz" is listed in the latest.json->assets (name) array before attempting to download it.

I have those changes stashed somewhere if you think that's a more reasonable approach than simply failing with a 404 when attempting to download the archive?

src/VersionUpdater/Installer.php
gefd: Thanks for catching this. I've pushed a fix that also addresses some incorrect `tempnam()` usage in the download classes!

bobthecow:
> I have those changes stashed somewhere if you think that's a more reasonable approach than simply failing with a 404 when attempting to download the archive?

honestly, i don't think a 404 is the end of the world. future us can deal with the bug reports when the available flavors change :)

src/VersionUpdater/SelfUpdate.php
```diff
@@ -0,0 +1,192 @@
...
+     *
+     * @param string $latestVersion
+     *
+     * @return string
+     */
+    private function getAssetUrl(string $latestVersion): string
+    {
+        $versionPostfix = '';
+        if (\strpos(Shell::VERSION, '-')) {
+            $versionPostfix = \substr(Shell::VERSION, \strpos(Shell::VERSION, '-'));
```
bobthecow, line 87: oh, sorry i missed this before. the version number looks like this: `v0.11.8-64-ga0c5d077+php70`. this is:

```
{version}-{git tag annotation}+{build flavor}
```

this lines up with semver's

```
{version}-{pre-release}+{build}
```

… unfortunately i don't name my release tarballs that way, so we need to compare/extract in one format and download in another 😬

gefd: Oh, my bad - I see I have the format of `Shell::VERSION` wrong when a build-flavor is present `{version}+{build}` rather than `{version}-{build}` and the url should always use `{version}-{build}`. 

I'm not sure I understand how the `-{pre-release}` part fits in the picture though. Does that need to be handled when using the `Shell::VERSION` value here? Or is the issue simply dealing with the `+` vs `-` in the case the version includes the build flavor?

bobthecow: Oh, you can ignore the pre-release bit. I was just explaining why you might have seen a `-*` in a version string. All we need to account for here is whatever comes after a `+`.

bobthecow (approved):

bobthecow:
Thank you so much for all your work on this!

Issue #746, opened 2022-10-24
Symfony deprecation trigger


Nek-:
Hello,

I was wondering if you have a solution to avoid having all of this in the terminal:

```
2022-10-24T22:05:16+00:00 [info] User Deprecated: Method "PhpParser\NodeVisitor::enterNode()" might add "int|Node|null" as a native return type declaration in the future. Do the same in implementation "Psy\CodeCleaner\UseStatementPass" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-10-24T22:05:16+00:00 [info] User Deprecated: Method "PhpParser\NodeVisitor::leaveNode()" might add "int|Node|array|null" as a native return type declaration in the future. Do the same in implementation "Psy\CodeCleaner\UseStatementPass" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-10-24T22:05:16+00:00 [info] User Deprecated: Method "PhpParser\NodeVisitor::beforeTraverse()" might add "?array" as a native return type declaration in the future. Do the same in implementation "Psy\CodeCleaner\NamespacePass" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-10-24T22:05:16+00:00 [info] User Deprecated: Method "PhpParser\NodeVisitor::enterNode()" might add "int|Node|null" as a native return type declaration in the future. Do the same in implementation "Psy\CodeCleaner\AbstractClassPass" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-10-24T22:05:16+00:00 [info] User Deprecated: Method "PhpParser\NodeVisitor::leaveNode()" might add "int|Node|array|null" as a native return type declaration in the future. Do the same in implementation "Psy\CodeCleaner\AbstractClassPass" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-10-24T22:05:16+00:00 [info] User Deprecated: Method "PhpParser\NodeVisitor::enterNode()" might add "int|Node|null" as a native return type declaration in the future. Do the same in implementation "Psy\CodeCleaner\AssignThisVariablePass" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-10-24T22:05:16+00:00 [info] User Deprecated: Method "PhpParser\NodeVisitor::beforeTraverse()" might add "?array" as a native return type declaration in the future. Do the same in implementation "Psy\CodeCleaner\CalledClassPass" now to avoid errors or add an explicit @return annotation to suppress this message.
[...]

PR #749, opened 2022-11-06
Add return types in doc to avoid symfony deprecation

Nek-:
This fixes https://github.com/bobthecow/psysh/issues/746 !

If I made a mistake in the code style or something please do not hesitate to highlight it.


bobthecow:
Awesome, thanks!

PR #753, opened 2023-01-23
[Fix] Docblock::tag(...) throws TypeError

ExeQue:
The `Docblock::tag(...)` method throws a TypeError when specifying a non-existent tag.

Fixed the method to allow for a `null` return alongside the normal `array` as that is what the intentions of the code suggests.

### DIFF
--- a/src/Util/Docblock.php
+++ b/src/Util/Docblock.php
@@ -212,9 +212,9 @@ class Docblock
      *
      * @param string $tag
      *
-     * @return array
+     * @return array|null
      */
-    public function tag(string $tag): array
+    public function tag(string $tag): ?array
     {
         return $this->hasTag($tag) ? $this->tags[$tag] : null;
     }


bobthecow:
Thanks for the fix! Unfortunately the lowest versions of PHP that we still support can't use nullable return types. They'll be dropped for v0.12, but in the meantime the correct fix is to remove the return type annotation and just update the docblock annotation.

I'm curious where you saw the actual error. As far as I can tell the only place that calls this in the PsySH codebase checks `hasTag` before calling anyway.

ExeQue:
I'm using the `Docblock` class in another context to verify that certain tags has a proper value and/or the docblock has a description (not at all in a way that's intended).

I've fixed my implementing code to use `hasTag` before attempting `tag` - I just found it suiting to report and fix the "bug" for this method since I happened to stumble across it.

bobthecow:
No, your fix here is the correct fix, I was just curious how you'd happened to notice the issue. Makes sense.

ExeQue:
I've fixed the method to allow any return types and added a TODO-comment for later reference in case anyone wonders why all the other methods have declared return types.

bobthecow:
Thanks again :)

PR #763, opened 2023-05-17
Use getenv() to fetch environment variable when absent from $_SERVER

supriyo-biswas:
At present, PsySH checks for environment variables in $_SERVER. As a result, when using PsySH from the `cli-server` SAPI, it may fail to look up the value of $HOME and thereby try to create the `~/.config/psysh` directory in `/` (which fails unless the user is root).

This PR fixes the issue by falling back on `getenv()` to find out the value of the environment variable.

### DIFF
--- a/src/ConfigPaths.php
+++ b/src/ConfigPaths.php
@@ -34,7 +34,8 @@ class ConfigPaths
     public function __construct(array $overrides = [], EnvInterface $env = null)
     {
         $this->overrideDirs($overrides);
-        $this->env = $env ?: new SuperglobalsEnv();
+
+        $this->env = $env ?: (\PHP_SAPI === 'cli-server' ? new SystemEnv() : new SuperglobalsEnv());
     }
 
     /**
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -147,6 +147,8 @@ class Configuration
             $this->configFile = $config['configFile'];
         } elseif (isset($_SERVER['PSYSH_CONFIG']) && $_SERVER['PSYSH_CONFIG']) {
             $this->configFile = $_SERVER['PSYSH_CONFIG'];
+        } elseif (\PHP_SAPI === 'cli-server' && ($configFile = \getenv('PSYSH_CONFIG'))) {
+            $this->configFile = $configFile;
         }
 
         // legacy baseDir option
--- a/src/SystemEnv.php
+++ b/src/SystemEnv.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of Psy Shell.
+ *
+ * (c) 2012-2023 Justin Hileman
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Psy;
+
+/**
+ * Environment variables implementation via getenv().
+ */
+class SystemEnv implements EnvInterface
+{
+    /**
+     * Get an environment variable by name.
+     *
+     * @return string|null
+     */
+    public function get(string $key)
+    {
+        if (isset($_SERVER[$key]) && $_SERVER[$key]) {
+            return $_SERVER[$key];
+        }
+
+        $result = \getenv($key);
+
+        return $result === false ? null : $result;
+    }
+}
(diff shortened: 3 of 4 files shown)


supriyo-biswas:
@GrahamCampbell, thank you for your comment.

While the [C function spec](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/) does say that the function need not be thread-safe, this change would make PsySH would only read the environment variables at the startup phase for its configuration.

In any case, what would you recommend in place of `getenv()`? The `cli-server` SAPI does not have `$_SERVER["HOME"]`, and neither does `$_ENV`.

bobthecow:
> this change would make PsySH would only read the environment variables at the startup phase for its configuration.

This isn't necessarily true. The `SuperglobalsEnv` class can be used at any time, not just during startup. Looking through current usage, other than in testing it's basically used by `ConfigPaths`, and _most_ things are initialized when starting PsySH before the first time _PsySH itself_ forks. The one clear culprit would be the `doc` command. We lazily instantiate the PDO instance for the database after you try running the `doc` command, so that's usually happening in a thread.

But! There's nothing stopping PsySH from being called initially from inside a non-threadsafe environment. You can just … start using PsySH from any userland code.

Maybe a better approach would be another Env instance specifically for `cli-server` SAPI, if it needs to be a special snowflake?

supriyo-biswas:
I special-cased out the `getenv()` logic to another class, and only use it when `PHP_SAPI === 'cli-server'`, let me know if that works.

supriyo-biswas:
> Instead of using _putenv_, you should be writing to the server superglobal.

I assume you meant to write "getenv"; regardless, to confirm, are you asking me to copy everything to the `$_SERVER`/`$_ENV` superglobal so that `SuperglobalsEnv()` can pick them up? Doesn't it introduce an unintended side effect for something that is meant to `get()` an environment variable?

GrahamCampbell:
getenv and putenv should essentially never be used. $_SERVER is where env variables are meant to live. PHP did a bad job of naming things. :trollface:

bobthecow:
This isn't because of putenv abuse. It's because the `cli-server` SAPI doesn't put things in superglobals like it should:

Screenshot 2023-05-18 at 11 30 11 AM

Screenshot 2023-05-18 at 11 30 53 AM

Screenshot 2023-05-18 at 11 31 17 AM

supriyo-biswas:
@GrahamCampbell 

I appreciate your reservations regarding the use of `get/putenv`, and I acknowledge that you have reservations about it. However, considering the current situation, I am unable to identify an alternative approach to address this problem.

It would be immensely helpful if you could kindly redirect the discussion towards finding a mutually agreeable solution.

bobthecow:
If we go this route, there are a couple more `$_SERVER` uses that we should pick up:

```patch
diff --git a/src/Configuration.php b/src/Configuration.php
index cac00c77..5bea4f37 100644
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -147,6 +147,8 @@ class Configuration
             $this->configFile = $config['configFile'];
         } elseif (isset($_SERVER['PSYSH_CONFIG']) && $_SERVER['PSYSH_CONFIG']) {
             $this->configFile = $_SERVER['PSYSH_CONFIG'];
+        } elseif (\PHP_SAPI === 'cli-server' && ($configFile = \getenv('PSYSH_CONFIG'))) {
+            $this->configFile = $configFile;
         }
 
         // legacy baseDir option
diff --git a/src/functions.php b/src/functions.php
index 10aaa02d..6daeed1b 100644
--- a/src/functions.php
+++ b/src/functions.php
@@ -155,6 +155,9 @@ if (!\function_exists('Psy\\info')) {
 
         $config = $lastConfig ?: new Configuration();
         $configEnv = (isset($_SERVER['PSYSH_CONFIG']) && $_SERVER['PSYSH_CONFIG']) ? $_SERVER['PSYSH_CONFIG'] : false;
+        if ($configEnv === false && \PHP_SAPI === 'cli-server') {
+            $configEnv = \getenv('PSYSH_CONFIG') ?: false;
+        }
 
         $shellInfo = [
             'PsySH version' => Shell::VERSION,
```

bobthecow:
@GrahamCampbell does this make sense? can you come up with a better way to resolve it?

GrahamCampbell:
Fine by me.

bobthecow:
Thanks @supriyo-biswas!

PR #765, opened 2023-07-26
Fix undefined array key when argv doesn't exists

BafS:
In some cases `$_SERVER['argv']` may not be set (I hit this issue with a project).


bobthecow:
Good catch. We handle the missing argv case everywhere else, but this code is a ported and minimally modernized portion of a legacy library and we must have missed it when cleaning that up.

Thanks!

PR #770, opened 2024-01-12
Avoid internal call to deprecated getVersion() method

Firehed:
After updating to v0.12.0, I get a deprecation warning emitted when running the shell. I have mine set up in a light Doctrine Console wrapper, though I suspect that's not relevant.

I tested this patch in my local vendor code and saw no negative impact, and it removed the deprecation warning.

### DIFF
--- a/src/Shell.php
+++ b/src/Shell.php
@@ -1509,7 +1509,7 @@ class Shell extends Application
      */
     protected function getHeader(): string
     {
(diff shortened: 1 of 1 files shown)


bobthecow:
Ahh, good catch. [The deprecation error is triggered but suppressed](https://github.com/bobthecow/psysh/blob/main/src/Shell.php#L1522), so unless you've set up your own handler via `set_error_handler` elsewhere (like your light Doctrine Console wrapper) it won't do anything at that point of execution.

PR #817, opened 2024-08-29
Fix missing arg for getMatches()

tareqas:


tareqas:
I need the input exactly as it is, including the same spacing, especially the ending space. Is there any way I can achieve this?

tareqas:
I want to determine if it's a partial or complete token by ending space. Explaining the implications using code would be easy.
I will share the code if I can develop something useful from it.

tareqas:
@bobthecow I've created this package using this feature. Also, I utilized https://github.com/bobthecow/psysh/pull/819 too 

https://github.com/tareqas/psym/blob/v1.0.1/src/TabCompletion/Matcher/SqlDqlMatcher.php

PR #819, opened 2024-09-04
Add support for non-combinable matcher

tareqas:


bobthecow:
I _think_ the reasons for this are resolved with the new experimental readline's completion engine!

PR #820, opened 2024-09-09
Confirm `less` is link before using it as such

piurafunk:
The PHP [docs](https://www.php.net/manual/en/function.readlink.php) of `readlink()` indicate that if the path provided is not a link, it fails.

> Note: The function fails if path is not a symlink, except on Windows, where the normalized path will be returned.

I am seeing this in Docker with PHP FPM image, with Alpine 3.20.3, PHP 8.2.23:

```
65aefc8b7840:/var/www/html$ php artisan tinker

   ErrorException 

  readlink(): Invalid argument

  at vendor/psy/psysh/src/Configuration.php:1266
    1262▕                 // n.b. The busybox less implementation is a bit broken, so
    1263▕                 // let's not use it by default.
    1264▕                 //
    1265▕                 // See https://github.com/bobthecow/psysh/issues/778
  ➜ 1266▕                 $link = @\readlink($less);
    1267▕                 if ($link !== false && \strpos($link, 'busybox') !== false) {
    1268▕                     return false;
    1269▕                 }
    1270▕ 

      +16 vendor frames 
  17  artisan:31
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -1263,7 +1263,7 @@ class Configuration
                 // let's not use it by default.
                 //
                 // See https://github.com/bobthecow/psysh/issues/778
-                $link = @\readlink($less);
+                $link = @\is_link($less) ? @\readlink($less) : $less;
                 if ($link !== false && \strpos($link, 'busybox') !== false) {
                     return false;
                 }


piurafunk:
Just discovered that the reason `less` is not a symlink in my environment is because we install it specifically in our Dockerfile:

```
RUN apk add --no-cache \
    ...
    less \
    ...
    ;
```

which then clobbers the symlink:

```
65aefc8b7840:/var/www/html$ ls -lah /usr/bin/less
-rwxr-xr-x    1 root     root      167.9K Apr 17 02:57 /usr/bin/less
```

Would still love this to be merged, though; it'll help stabilize in different environments.

piurafunk:
Yes, I imagine that is Laravel. In combination with PHP 8, the `@` symbol reports it's error level, see [docs](https://www.php.net/manual/en/language.operators.errorcontrol.php). Which causes Laravel to throw a new ErrorException.

It seems like defensive programming to first check if it is a link before treating it as such. Do you agree with that?

piurafunk:
I've updated that in 0c59c2de22af1f63fee5d9d2d59c70e26f24b856.

piurafunk:
@bobthecow How soon/frequently do you do releases? I'd love to upgrade to get this change in our app.

PR #840, opened 2024-11-29
Tighten up types and initialization for private properties and methods.

bobthecow:
- Remove docblock param annotations that don't contribute anything
 - Tighten up param and return types for private methods.
 - Move a bunch of private static variables to private constants.
 - Ensure that commands only call shell-y things on shell instances
 - Tighten up types and initialization for private properties
 - Misc type cleanup
 - Remove (now unsupported) phpstan config option.

_n.b. This doesn't change public or protected property or method types. Those can be cleaned up for the next major version release._

_It **does** cause PHP to throw for some uninitialized-but-required properties (e.g. those provided via interface injection). These aren't backwards compatibility breaks, because they were broken before, just with a less-useful exception._

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -80,58 +80,57 @@ class Configuration
         'yolo',
     ];
 
-    private $defaultIncludes;
-    private $configDir;
-    private $dataDir;
-    private $runtimeDir;
-    private $configFile;
-    /** @var string|false */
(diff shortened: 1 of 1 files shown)


mcdruid:
This looks fantastic - it's going to be much harder to get up to Gadget Chain-type mischief after these changes!

Nice work @bobthecow!

PR #868, opened 2025-06-29
Avoid matching functions against ->

igorsantos07:
I'm sorry if my train of thought is a bit limited, but it doesn't make sense to call a function after `->«any_partial_string»`.... does it? 👀


bobthecow:
Makes sense, thanks!

PR #870, opened 2025-07-04
Use https urls

bebehr:
Changes:

- Switched from http://psysh.org to HTTPS.

- http://justinhileman.com does not support HTTPS. Removed url, only content there is smiley :)


bobthecow:
Thanks!

Issue #888, opened 2025-10-30
Incorrect output for arrays


daniser:
PHP version: **8.4.1**
Laravel: **12.36.1**
Tinker: **2.10.1**

PsySH 0.12.12:

![Image](https://github.com/user-attachments/assets/573b4159-8e88-4040-8e67-72152e6c6142)

PsySH 0.12.13 and 0.12.14:

![Image](https://github.com/user-attachments/assets/f864b5fe-e18b-4324-983d-41afd607665e)

Collection output is identical, the issue is only with arrays.
P.S. What's with the `The system cannot find the path specified` message?

vagkaefer:
Related with https://github.com/bobthecow/psysh/issues/512

bobthecow:
Thank you for opening this issue. I'm going to close it to consolidate conversation in #512

Issue #512, opened 2018-07-09
Show short dump on commands which do not look like they were meant for inspection


tgr:
Object dumps are great for inspecting state but when I am just setting up the state and performing a series of commands, each of which returns a complex object with a multi-page dump, that gets annoying fast. It would be nice if the interpreter would recognize which commands are not meant for inspection (commands including an assignment operator would be a decent first approximation) and use a short version of the dump (probably just the class name would work well enough) in that case.

#218 is sort of related.

phil-s:
You can follow the command with another command. PsySH only prints the value of the last statement it executed. e.g.:

    >>> thing_which_returns_a_lot_of_state();1;
    1

tgr:
> Is there currently a way to re-print the most recent return value?

Yeah, `dump $_`. But the more intuitive (to me) version is

    $foo = someCommand();
    $foo

where the second command can't have any other possible reason than inspection. Similarly a command like `$foo['bar'][$baz]` is no doubt used for inspection. So either using a full dump when there is no assignment, or using it when there is no assignment and no function/method call (ie. no side effect) feels natural to me.

> You can follow the command with another command.

Thanks, that's a nice workaround.

bobthecow:
The implicit dump for return values is _supposed_ to be printing a short dump. The `dump` command is there if you want more info. If it's not doing that now, we should fix it :)

bobthecow:
This … took longer than it should have :-/ 

As of the latest commit in `@dev`, we're using a more abbreviated format for return values of things that look like taking an action, and a _slightly more verbose_ output for things that look like inspection:

<img width="786" height="966" alt="Image" src="https://github.com/user-attachments/assets/62b346d0-bb6e-45f2-b0b7-049af6096619" />

We can twiddle the knobs after living with it to see where it makes sense to draw the lines, but hopefully this improves things a bit!

jasonmccreary:
@bobthecow, hey, thanks for the package. Use it daily within my Laravel app. To that point, this changed the "expected" behavior pretty significantly. I definitely understand the motivation from the OP. But given its age, your noted effort, and the impact to large downstream projects like Laravel, I wonder if it is necessary.

Maybe instead this could be opted into. Like a per session flag or an ENV. Especially as you "twiddle the knobs".

Thanks!

bobthecow:
@jasonmccreary thanks for the response! do you mind opening an issue with feedback about the change, so we can improve the defaults for everyone, rather than just "please don't"? 😉

[...]

PR #891, opened 2025-11-14
feat: create configuration option to enable/disable the consise retur…

vagkaefer:
# Concise Return Value Configuration

## Overview

Both I and other users (mainly Laravel users) experienced discomfort with the change made in commit 932daa0, linked to issues #512 and #888.

So, instead of just complaining and debating, I created an option for each user to choose how they want the default output to be. However, the concise mode, as implemented in the aforementioned commit, remains the default.

With this changes, PsySH now includes a configurable option to control whether return values are displayed in a concise format for "action" commands (like assignments and setters) versus "inspection" commands (like variable reads and getters).

## Configuration Option

- **Option name**: `useConciseReturnValue`
- **Type**: `boolean`
- **Default**: `true` (enabled)

## Behavior

When **enabled** (default):
- Action commands (assignments, setters, etc.) show abbreviated output
- Inspection commands (variable reads, getters, etc.) show full detailed output

When **disabled**:
- All commands show full detailed output regardless of type
### Option 1: Via Configuration File

Create or edit your PsySH configuration file (usually `~/.config/psysh/config.php`):

```php
<?php

return [
    // Always show full output for all commands
    'useConciseReturnValue' => false,

    // OR: Enable concise output for actions (default behavior)
    // 'useConciseReturnValue' => true,
];
```

### Option 2: Programmatically

```php
<?php

use Psy\Configuration;
use Psy\Shell;

$config = new Configuration([
    'useConciseReturnValue' => false, // Show full output always
]);

$shell = new Shell($config);
$shell->run();
```

### Option 3: At Runtime

```php
$config = new Configuration();
$config->setUseConciseReturnValue(false); // Disable concise output
```

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -41,19 +41,19 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 class Configuration
 {
-    const COLOR_MODE_AUTO = 'auto';
-    const COLOR_MODE_FORCED = 'forced';
-    const COLOR_MODE_DISABLED = 'disabled';
+    public const COLOR_MODE_AUTO = 'auto';
+    public const COLOR_MODE_FORCED = 'forced';
+    public const COLOR_MODE_DISABLED = 'disabled';
 
-    const INTERACTIVE_MODE_AUTO = 'auto';
-    const INTERACTIVE_MODE_FORCED = 'forced';
-    const INTERACTIVE_MODE_DISABLED = 'disabled';
+    public const INTERACTIVE_MODE_AUTO = 'auto';
+    public const INTERACTIVE_MODE_FORCED = 'forced';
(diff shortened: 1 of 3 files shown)


nickjbedford:
Great! I have been trying to figure out why this was happening in the last few weeks as a Laravel Tinker user. Please pull this change as me and probably a lot of developers rely on the automatic outputting of the results of "action" expressions instead of having to now store them in a variable then enter _another_ expression just to see the result.

bobthecow:
@vagkaefer I appreciate you taking the time to open this PR!

I don't think just adding a configuration to opt-out is the right answer. See my comment in the original issue for more thoughts: https://github.com/bobthecow/psysh/issues/512#issuecomment-3570622341 and please join us over there with yours!

bobthecow:
Thanks again. The change has been reverted for now 🙂

PR #895, opened 2025-11-25
Don't call curl_close() on PHP >= 8.0.0

mpesari:
`curl_close()` is noop since 8.0.0 and emits `E_DEPRECATED` since 8.5.0.

This changes the code to call `curl_close()` only on PHP < 8.0.0.

* https://www.php.net/curl_close
* https://github.com/php/php-src/blob/php-8.0.0/UPGRADING#L1016-L1020
* https://github.com/php/php-src/blob/php-8.5.0/UPGRADING#L416-L418


bobthecow:
Thank you!

I'll merge this in a bit, but I'm going to use your PR to debug code coverage reports first 😛

bobthecow:
Thanks again, this will go in the next release!

PR #944, opened 2026-06-10
Fix error message formatting matching too many characters

### LINKED ISSUE #943: Exception formatting seems to be broken
The new regex to remove the `.../psysh/src/...` caller path when we get an error seems to match to many characters.

```php
        $message = \preg_replace(
            [
                "#(?:[A-Za-z]:)?[\\\\/][^\\r\\n]*?[\\\\/]src[\\\\/]Execution(?:Loop)?Closure\\.php\\(\\d+\\) : eval\\(\\)'d code#",
                "#\\bsrc[\\\\/]Execution(?:Loop)?Closure\\.php\\(\\d+\\) : eval\\(\\)'d code#",
            ],
            "eval()'d code",
            $message
        );
```
The `[^\\r\\n]` part of the first pattern end up matching whitespaces so the error messages matches up until the first `\` or `/`.

for example: 
```
X\Y::function(): Argument #1 ($values) must be of type array, string given, called in  /Z/vendor/psy/psysh/src/ExecutionLoopClosure.php(55) : eval()'d code on line 1
```
becomes 
```
Xeval()'d code on line 1
```

bilboque:
Closes #943 

Matches until the first non whitespaces character to extract the  `psysh/src` path of the error message

### DIFF
--- a/src/Shell.php
+++ b/src/Shell.php
@@ -1977,7 +1977,7 @@ class Shell extends Application
 
         $message = \preg_replace(
             [
-                "#(?:[A-Za-z]:)?[\\\\/][^\\r\\n]*?[\\\\/]src[\\\\/]Execution(?:Loop)?Closure\\.php\\(\\d+\\) : eval\\(\\)'d code#",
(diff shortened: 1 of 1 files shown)


bobthecow:
Thank you!

PR #951, opened 2026-08-09
Make include loading exception-safe

binaryfire:
PsySH currently stops loading configured include files when one of them has a parse error. Later includes never run, and a failure while loading or reporting can leave PsySH's temporary error handler active.

This change reports PHP errors as well as exceptions, continues loading the remaining files, and always restores the previous error handler.

`include_once` and scope behavior are unchanged. Tests cover parse errors, continued loading, scope precedence, and error-handler cleanup.

### DIFF
--- a/src/Shell.php
+++ b/src/Shell.php
@@ -830,10 +830,8 @@ class Shell extends Application
 
     /**
      * Load user-defined includes.
-     *
-     * The shell output must be configured before include failures can be reported.
      */
-    public function loadIncludes(): void
+    private function loadIncludes()
     {
         // Load user-defined includes
         $load = function (self $__psysh__) {
--- a/src/Shell.php
+++ b/src/Shell.php
@@ -830,20 +830,25 @@ class Shell extends Application
 
     /**
      * Load user-defined includes.
+     *
+     * The shell output must be configured before include failures can be reported.
      */
-    private function loadIncludes()
+    public function loadIncludes(): void
     {
         // Load user-defined includes
         $load = function (self $__psysh__) {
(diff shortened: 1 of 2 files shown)


bobthecow:
Thanks for the PR!

> It also makes loadIncludes() public so integrations can use the same behavior without reimplementing it.

Why, though? This is kind of a weird thing to make a public API.

binaryfire:
Hi @bobthecow.

> Why, though? This is kind of a weird thing to make a public API.

This came from Laravel Tinker’s `--execute` path. Tinker calls `setIncludes()`, then calls `Shell::execute()` directly. `execute()` doesn’t load those files; that only happens inside `run()`. As a result, `php artisan tinker bootstrap.php --execute='...'` currently ignores bootstrap.php. 

I considered using the non-interactive `run()` path but it isn’t equivalent to `--execute`. It sends the input through PsySH’s command handling and automatically writes the return value. Tinker’s current path evaluates the string as PHP and doesn’t print its return value.

I made `loadIncludes()` public so Hypervel’s Tinker integration can load the configured files before executing the code without copying PsySH’s include logic. Hypervel is a Swoole framework I’m co-creating.

That said, I agree it’s a pretty low-level public API. Happy to drop the visibility change from this PR and handle the integration API separately, or change it to a higher-level API if there’s another approach you’d prefer.

bobthecow:
got it. let's split that out from the exception handling, and figure out what a clean extension point is in another PR?

binaryfire:
@bobthecow Done

bobthecow:
Thanks!

PR #952, opened 2026-08-09
Restore signal state after shell execution

binaryfire:
When code runs through `Shell::execute()`, PsySH sets up Ctrl-C handling like it does in the REPL. That state needs to be restored afterwards, including when user code throws.

This PR adds an `afterExecute()` listener hook paired with `onExecute()`. `SignalHandler` uses it to restore the previous SIGINT handler and async-signal mode. Nested executions share the outer signal state, so only the outer call restores it.

It also tracks nested `run()` calls with a counter so an inner run can’t clear the outer run’s active state. `afterLoop()` stays paired with `beforeLoop()` and keeps its existing loop-only behavior.

### DIFF
--- a/src/ExecutionClosure.php
+++ b/src/ExecutionClosure.php
@@ -27,37 +27,44 @@ class ExecutionClosure
     {
         $this->setClosure($__psysh__, function () use ($__psysh__) {
             try {
-                // Restore execution scope variables
-                // @phan-suppress-next-line PhanTypeNonVarPassByRef assigning to a temp variable pollutes scope
-                \extract($__psysh__->getScopeVariables(false));
+                try {
+                    // Restore execution scope variables
+                    // @phan-suppress-next-line PhanTypeNonVarPassByRef assigning to a temp variable pollutes scope
+                    \extract($__psysh__->getScopeVariables(false));
 
-                // Buffer stdout; we'll need it later
-                \ob_start([$__psysh__, 'writeStdout'], 1);
+                    // Buffer stdout; we'll need it later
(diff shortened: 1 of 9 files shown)


bobthecow:
Can we use the execution depth from the previous PR here?

binaryfire:
@bobthecow Done. I’ve updated it to use the execution depth from #954. The separate run flag is still needed so a direct `execute()` call doesn’t change terminal settings.

bobthecow:
Thank you. The signal snapshot/restore work and test coverage here look good, but I'm not sure `afterLoop()` is the right place for it. That's a public loop-listener lifecycle hook, so using it for execution cleanup creates unpaired callbacks and misses command boundaries such as `timeit`, and would technically be a BC break.

I think the cleanest thing would be to keep your signal-state implementation but move restoration into an execution-specific `afterExecute()` hook, paired with `onExecute()`. We can also turn `runActive` into a nesting counter. I’m happy to push a refactoring commit on top of your branch, if that works for you!

binaryfire:
@bobthecow Sure, no problem. I just realised I missed that `afterLoop()` is paired with `beforeLoop()`, not `onExecute()`. I'd like to give this one more pass, if that's cool with you. Then after that please feel free to make any changes directly.

binaryfire:
@bobthecow I’ve pushed an updated implementation. Please feel free to make any additional changes.

bobthecow:
Sorry about the delay @binaryfire. I've been moving :)

This is merged, and I added one more stacked change that cleans up a couple more edge cases. Wanna rebase #953?

binaryfire:
@bobthecow No worries! Yes, will do.

PR #953, opened 2026-08-09
Restore shell state after full runs

binaryfire:
A PsySH session can skip listener cleanup if an include, listener, or command throws. That can leave signal handling and non-interactive mode in the wrong state.

This PR makes cleanup run for every session that starts successfully, without losing the exit code. It also keeps nested ProcessForker executions from restoring signal state too early, and restores the parent’s original SIGINT handler and async-signal setting when the child exits.

### DIFF
--- a/src/Completion/TypeResolver.php
+++ b/src/Completion/TypeResolver.php
@@ -850,7 +850,11 @@ class TypeResolver
             return [];
         }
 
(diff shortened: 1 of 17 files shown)


binaryfire:
I'll update this after https://github.com/bobthecow/psysh/pull/952 merges

binaryfire:
@bobthecow I’ve updated the PR to work with the latest changes on main.

bobthecow:
Thanks!

PR #954, opened 2026-08-25
Load configured includes in execute()

binaryfire:
Laravel and Hypervel Tinker set include files on the shell before calling `execute()`, but PsySH only loads them when `run()` is used. Switching to `run()` isn’t equivalent because it treats input as PsySH commands and prints return values.

This makes `execute()` load configured includes itself. They’re loaded once for the outer call, so calls made while PsySH is already running don’t load them again.

### DIFF
--- a/src/Shell.php
+++ b/src/Shell.php
@@ -830,8 +830,10 @@ class Shell extends Application
 
     /**
      * Load user-defined includes.
+     *
+     * The shell output must be configured first; otherwise, reporting an include failure will abort loading.
      */
-    private function loadIncludes()
+    protected function loadIncludes()
     {
         // Load user-defined includes
         $load = function (self $__psysh__) {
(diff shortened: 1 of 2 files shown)


bobthecow:
thanks!

i’m not sure `loadIncludes()` is the right extension point here. it still requires integrations to subclass `Shell` and know the correct output/boot/include/execute lifecycle ordering.

is there a reason we'd want to execute _without_ honoring configured includes? what if we keep it private and have `execute()` load configured includes when it’s the outermost execution? we’d need to track nesting so inner calls from `run()`, `timeit`, reflection commands, etc. don’t load them again. that would make `setIncludes()` work consistently and fix the Laravel/Hypervel type cases without requiring a custom shell subclass.

binaryfire:
@bobthecow Done. Could you please check it out when you have a moment and let me know what you think?

bobthecow:
Looks great, thank you!

PR #955, opened 2026-09-06
Resolve a generic docblock type to its container

o-kima:
`@return Collection<int, User>` describes a `Collection`, but `parseDocblockTypes()` resolves the type *arguments* rather than the type itself, so completion offers the members of the wrong class.

```php
if (\preg_match('/^[\w\\\\]+<(.+)>$/', $typeString, $matches)) {
    return $this->parseDocblockTypes($matches[1], $classContext);
}
```

With more than one argument it produces a name no class ever has. The whole argument list is passed on as a single type, so `Collection<int, User>` resolves to a class called `"int, User"`; nothing downstream can reflect on it, and no members are offered at all. That's the common shape in Laravel — `Collection<int, User>`, `Collection<TKey, TValue>` — so on a Laravel model the effect is that a chain simply stops.

It's also the opposite of what the branch two lines above does for `User[]`, which resolves to nothing on the grounds that an array has no members to complete. A `Collection` does have members — they're the Collection's.

Resolving the outer name settles all three cases:

| `@return` | before | after |
| --- | --- | --- |
| `Collection<int, User>` | `"int, User"` | `Collection` |
| `Collection<User>` | `User` | `Collection` |
| `array<string, User>` | `"string, User"` | nothing, matching `User[]` |
| `User[]` | nothing | nothing |

Four cases added to `methodChainingProvider`, including one that chains off the collection (`$repo->all()->first()`) to show the members now resolve. All four fail on `main` and pass with the change; the full suite is green (4792 tests).

I ran into this building browser completion on `CompletionEngine` for [spatie/laravel-web-tinker](https://github.com/spatie/laravel-web-tinker/pull/125) — thanks for the pointer to the engine, it's a much better base than what I had.

One thing I wasn't sure about: if the argument unwrapping is deliberate for some case I haven't found, the alternative would be returning both the container and its last argument. I went with the container alone, since that's what the expression evaluates to and it matches the `[]` branch. Happy to change it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MeokeyBuZ3JKQYJ5L6gscZ

### DIFF
--- a/src/Completion/TypeResolver.php
+++ b/src/Completion/TypeResolver.php
@@ -850,7 +850,11 @@ class TypeResolver
             return [];
         }
 
-        if (\preg_match('/^[\w\\\\]+<(.+)>$/', $typeString, $matches)) {
+        // A generic names the type of the value itself, not of what it holds:
+        // `Collection<int, User>` is a Collection. Resolving the arguments
+        // instead offers the members of the wrong class, and for more than one
+        // argument produces a type name no class ever has ("int, User").
+        if (\preg_match('/^([\w\\\\]+)<.+>$/', $typeString, $matches)) {
             return $this->parseDocblockTypes($matches[1], $classContext);
         }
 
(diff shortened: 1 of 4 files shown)


bobthecow:
Thank you!

